PR #1, opened 2020-04-27
Get PI_PM test working

sgherbst:
This PR closes out issue DragonPHY issue 60 by fixing the second PI-related test that was broken, called ``PI_PM``.  This test varies the PI input codes while monitoring the PI phases using the built-in phase monitors.  Also as part of this commit, I updated the modeling approach for ``del_PI`` so that this block is now modeled at gate level, which includes special low-delay ``inv_PI`` inverters.  This cut the nominal gain of the PI in half, so I had to adjust the ``PI`` test accordingly.  Finally, I reformatted ``del_PI`` to make it easier to read.

### DIFF
--- a/tests/new_tests/PI/test.sv
+++ b/tests/new_tests/PI/test.sv
@@ -128,8 +128,8 @@ module test;
         #(10ns);
 
         // run desired number of trials
-        // TODO: explore behavior beyond 350
-        for (int i=0; i<=350; i=i+1) begin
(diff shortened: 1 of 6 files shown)


sgherbst:
@sjkim85 if you're planning to submit the ``analog_core`` PR today, I can review and merge that first

sgherbst:
Since this branch has to be merged with master anyway, I'm going to close the PR and re-submit one later today that includes the AC tests.

Issue #6, opened 2020-04-28
ENOB for ADC is low


sgherbst:
Simulated ENOB is currently ~4.85 for the AC test (Fstim=1.023GHz).  Could be a modeling, testing, or design issue, but it's important to log here because this is an important performance metric.

sjkim85:
I will look AC test results. I may be due to the obsolete setting of V2T.

sgherbst:
Thanks.  Two possible causes:
1.  I have attempted to align PI clocks, but the alignment may not be perfect.  If you have suggestions on the tuning method to get exact spacing of all 16 samples, I would be interested to hear them.  I think this is important for lab testing too.
2. TDC nominal delay, standard deviation, or jitter values may be out of date.

sjkim85:
1) The yaw that I did for the lab testing is the following 
Did the sine regression with data comes from each ADC slice (16 sinusoidal regression in total) and estimated phase of each regression. I tweaked the PI code until I get uniform phase spacing.

2) I think Td_inv_nom = 15p, std=1p, jit =0.1p is reasonable value. One more thing I did in the lab testing was I swept ctl_dcdl_coarse to get the better ENOB (delay step of the dcdl_coarse is roughly ~40ps)

sgherbst:
This was resolved awhile ago.

PR #11, opened 2020-04-29
Add AC, AC_REPLICA, and PI_PM tests

sgherbst:
This PR adds three new tests:
* AC: An ~1 GHz sine wave is sent into the ADC and sampled by all 16 slices in a time-interleaved fashion.  The ENOB is checked to be above 4.5.  A separate GitHub is has been opened to investigate why the ENOB is low (#6).  There are several hacks in the test due to other open issues, so we need to be certain to revisit this test in a week or so.
* AC_REPLICA: An ~10 MHz sine wave is sent into replica ADC inputs while checking that the ENOB of both channels is above 5.0.  Since the input frequency is low, this is not too different from the DC_REPLICA test.
* PI_PM: The phase of the PI channels is monitored using the internal phase monitor.

In addition, I reorganized the gate models into the subfolder ``new_cpu_models/analog_core/gates``.  Some gates from recent updates to ``analog_core`` still have to have models added.


sgherbst:
This PR is now passing and ready for review.

PR #67, opened 2020-05-05
add jtag register map for mdll and update some mdll rtls

bclim:
1. Added the jtag register map markdown for MDLL.
2. Some RTLs of the MDLL is revised after P&R


sgherbst:
@bclim looks like this branch is out-of-date with the ``master`` branch, would you mind merging the latest changes from master and pushing to this branch?  Thanks!

PR #73, opened 2020-05-05
Weight dev

zamyers:
1. Added the Weight Management Engine and a block level test for the WME
2. Added a bypass PI CTL functionality into the JTAG
3. Wired out the CDR's special features for detecting  (Input: Ramp Clock, Output: Freq Level Crossing)
4. Added second memory macro to support sampling the output of the FFE and MLSD. This has the same trigger functionality as the other macro, however it has a different JTAG interface (to simply the control).


sgherbst:
Thanks!  @zamyers is it alright if I merge #67 first since it arrived before this PR?  I don't think there should be any conflicts but you will have to merge changes from ``master`` if I do that.

zamyers:
That is fine. Just shoot me an email once you’ve merged it, I will pull the origin/master and re-push.

zamyers:
Should be good to merge now!

vlog/new_chip_src/digital_core/digital_core.sv
```diff
@@ -44,12 +49,41 @@ module digital_core import const_pack::*; (
...
+                assign trunc_est_bits[16+gj][gi] = checked_bits[gi + gj*8];
+            end
+        end
+    endgenerate
 
     wire logic [Npi-1:0] scale_value [Nout-1:0];
     wire logic [Npi-1:0] unscaled_pi_ctl [Nout-1:0];
     wire logic [Npi+Npi-1:0] scaled_pi_ctl [Nout-1:0];
 
+//    initial begin
```
sgherbst, line 76: See tests like ``new_tests/DC`` for a testing pattern that does not require adding waveform dumping commands to the design itself (basically, there is a define macro ``DUMP_WAVEFORMS`` that causes an SHM database to be dumped from the testbench, rather than the design; the macro is left undefined for regression runs to speed them up)

vlog/new_chip_src/digital_core/digital_core.sv
```diff
@@ -156,20 +190,47 @@ module digital_core import const_pack::*; (
...
         .pi_ctl(pi_ctl_cdr),
         .cdbg_intf_i(cdbg_intf_i)
     );
 
     genvar j;
     generate
         for (j=0; j<Nout; j=j+1) begin
-            assign scale_value[j]        = (ddbg_intf_i.en_ext_max_sel_mux ? ddbg_intf_i.ext_max_sel_mux[j] : adbg_intf_i.max_sel_mux[j])<<4;
+            assign scale_value[j]        = (((ddbg_intf_i.en_ext_max_sel_mux ? ddbg_intf_i.ext_max_sel_mux[j]: adbg_intf_i.max_sel_mux[j]) + 1)<<4) -1;
             assign unscaled_pi_ctl[j]    = pi_ctl_cdr[j] + ddbg_intf_i.ext_pi_ctl_offset[j];
```
sgherbst, line 203: Is it intended that ``unscaled_pi_ctl`` will overflow?  (Not sure whether wrapping is expected here.)  Since ``pi_ctl_cdr`` is 0-511 and so if ``ext_pi_ctl_offset``, the max value of ``unscaled_pi_ctl`` without wrapping would be 1,022, but the signal can only values 0-511.

Issue #61, opened 2020-05-01
synopsys translate_off


sgherbst:
Just noticed that this is used in at least two places to initialize values.  I think we need to be really careful to make sure that we're not relying on these initial values in the real design.  Here are the places where I found this:

https://github.com/StanfordVLSI/dragonphy2/blob/2d280b99c7884aedabea331e1e2498c6bb49c457/vlog/new_chip_src/analog_core/V2T_clock_gen.sv#L45-L49

https://github.com/StanfordVLSI/dragonphy2/blob/2d280b99c7884aedabea331e1e2498c6bb49c457/vlog/new_chip_src/analog_core/PI_local_encoder.sv#L36-L42

If possible, I suggest we remove these non-synthesizable initializations.

bclim:
Sometimes it is necessary if the counter reg doesn’t have reset. But it looks like those regs have an reset input. So, as Steven suggested, it would be good to get rid of the initial statement and reset the counter properly.
Another way (less conservative way) in general is to randomly initialize it using $urandom().

btw, I looked at the code, but I don’t quite understand the code to have two async resets, and I was wondering how that is mapped to a gate.

bclim:
That will be mapped to a flop with both set and reset. But some stdcell providers might not have that kind of flops

sgherbst:
Done

Issue #22, opened 2020-04-30
Remove clk_cdr and replace with a clock enable signal


sgherbst:
This task involves changes to both ``analog_core`` and ``digital_core``.  We want to remove ``clk_cdr`` and replace it with a clock enable signal that indicates when ``pi_ctl_cdr`` is valid and should be loaded.

sgherbst:
This was resolved awhile ago.

Issue #21, opened 2020-04-30
Review GLITCH test


sgherbst:
The expectation is that changes of +/- 1 PI code should never cause a glitch in the PI output.  I think the current GLITCH test only checks for glitches in steady-state, so this needs to be updated.

sgherbst:
I updated the test to do this, and found some glitches, so I'm re-assigning to @sjkim85.  To reproduce the issue, check out the ``glitch_tests`` branch and run ``pytest tests/new_tests/GLITCH -s``.  Two different kinds of errors are seen:
1. True glitches (i.e., very short pulses and multiple clock edges).  These only occur at a few specific code transitions, like 496 -> 495 and 240 -> 239.
2. Low-resolution issues, where certain code transitions cause a much bigger change in phase than expected (~10ps instead of ~1ps).  An example would be the 255 -> 256 transition.  This kind of error occurs much more frequently on negative code transitions than positive code transitions.  Also, it appears to have a periodic behavior while codes are decreasing (i.e., every "N" codes there is a big step)

sjkim85:
I was trying to see waveforms but I got this errors.
(I ran the pytest  in the glitch_tests branch as you suggested after I did "python make.py --view fpga)
[cid:35241fae-bd5f-43ea-b73d-0b9ec09f645b]
________________________________

sgherbst:
You can send me a Slack message or email with the error message and we can debug from there.  In general, I suggest re-running ``pip install -e .`` in the top-level directory to reinstall DragonPHY when you see errors, since there may be new package dependencies that were added.  Also, you'll need to switch to the ``pwl_cos`` branch on DaVE since there is an open pull request.  When in doubt, the definitive install steps can always be found in the ``regress.sh`` script.

sgherbst:
Note: while @sjkim85 is debugging this, I marked the GLITCH test "work in progress".  To run the GLITCH test, add ``--runwip`` to the normal arguments you pass to ``pytest`` (otherwise the test will be skipped).  Then, as a condition for resolving this issue, we need to remove the "work in progress" marker from the test.

sgherbst:
@sjkim85 Thanks for all of your work on this issue.  I noticed that it's still marked as "work-in-progress", meaning that the GLITCH test is not being run in the regression.  Could you update this in the next PR?
This can be done by changing ``pytest.mark.wip`` in ``test_glitch.py`` to:
```python
[...]

PR #74, opened 2020-05-06
Analog core2

sjkim85:
Updated analog_core associated with issue #3 , #21 , #22 , #61


sgherbst:
@sjkim85 Looks like the two failing tests are ``PB_ONLY`` and ``PI``.

``PB_ONLY`` is indicating there is some nonlinearity seen in the ``phase_blender`` model:
https://github.com/StanfordVLSI/dragonphy2/blob/35c71a2436ddf0583620c47ed2e969c43f5364f9/tests/new_tests/PB_ONLY/test_pb_only.py#L175-L176
It could be that this is intended (i.e., it's a more realistic model), or it could be an error in the model.  One of the ways to debug this is to look at the graphs produced in the ``PB_ONLY/build`` folder (``pb_ph.eps``, ``pb_freq.eps``, ``pb_duty.eps``) to see if they are as expected.

For ``PI``, it looks like the test would have passed, but an undeclared variable was used in one of the print statements:
https://github.com/StanfordVLSI/dragonphy2/blob/35c71a2436ddf0583620c47ed2e969c43f5364f9/tests/new_tests/PI/test_pi.py#L153
(The variable ``deltas`` has not been defined).  Perhaps there was a merge issue in making updates to the ``PI`` test?  Anyway, if you change ``deltas`` to ``np.diff(y)`` in that line, you should be good to go.

sjkim85:
1) Fixed analog_core for PI test and PB_ONLY test  
- For PI test, error in test bench (test_pi.py) is fixed (-->diff(y))
- For PB_ONLY test, testbench (test_pb_only.py) is modified to reflect more realistic scenario.
     --> input delay diff sweep range : +/- 125ps --> +/- 60ps (exclude zero input difference)
     --> add input ctl code retiming in test.sv

2) QTM models of analog_core/input_buffer/output_buffer/butterphy_top and script for generation them are added

sgherbst:
@sjkim85 looks like this is failing because ``analog_core`` is including the file ``/aha/sjkim85/github_repo/dragonphy2/inc/new_asic/iotype.sv``.  Can you change this to ``iotype.sv``?  By the way, in your synthesis scripts you can set the ``search_path`` variable in TCL so that DC can find the files.  ``new_tests/SYN`` has an example of this.

sgherbst:
Two other minor requests:
1. Can you change the ``DUMP_WAVEFORMS`` setting in the ``PI`` test to ``False``?  When waveform dumping is enabled the regression is slower.
2. Can you remove the ``.nfs*`` file from the ``new_chip_src/analog_core`` directory and add ``.nfs*`` to ``.gitignore``?
Thanks!

sjkim85:
Steven, 
can you review this and merge to master?
major changes about glitch are 
1) new_chip_arc/analog_core/phase_interpolator.sv 
2) new_cpu_models/analog_core/phase_blender.sv

sgherbst:
@sjkim85 Thanks for the updates.  It looks like a bunch of timing models (like ``n_and_arb_fixed``) were removed from new_cpu_models/analog_core/gates -- was this intended?  On a related note, there are some new gates like ``n_or_v2t_fixed`` that don't have timing models in ``new_cpu_models/analog_core/gates``.  In both cases, the result is that these gates are being simulated with zero delay, which could yield a misleading simulation result.

The update to the ``phase_blender`` model looks fine -- I think the key assumption is that the ``td`` value must be greater than the time difference between the two edges.  Given that the current ``td`` value is ``100ps``, I think that assumption is likely satisfied.

Last comment is minor -- it looks like ``DUMP_WAVEFORMS`` was enabled for the ``GLITCH`` and ``MUX4_GF_ONLY`` tests.  This can wait until a future PR to fix, but please do remember to update this since it slows down the regression run, meaning that PRs take longer to get approved.

sjkim85:
1) Yes it was intended. I needed to clean up the files since there were many modules and old/updated names are mixed. I deleted files which are not in use in current design and moved files from new_cpu_models to new_chip_src except modules that are required to be modeled.  
For example, inv_v2t_1_fixed.sv exists in new_chip_src only. Timing model of this cell is not necessary  because it is used only for V2T_clock_gen_S2D.sv which has timing model already.

2) Yes. td is always larger than the input time difference in current design (and will be in silicon)

3) I will fix that and push again

sjkim85:
I think n_and_arb_fixed.sv (for arbiter) needs timing model. I will add it.

PR #79, opened 2020-05-07
Add preliminary top-level synthesis flow

sgherbst:
This PR adds a preliminary top-level synthesis flow that uses mflowgen.  It includes a new BuildKite task called **test_mflowgen** that synthesizes ``dragonphy_top`` in **FreePDK45** (in runs in parallel with simulation and emulation tasks).  Because this is a superset of the ``SYN`` test, I removed ``SYN`` from the CPU simulation regression test set.

Synthesis is still a work in progress, and in particular the Quick Timing Model (QTM) for ``analog_core`` has to be updated; right now a Verilog stub with no timing information is used just to get the flow running.


sjkim85:
I will fix dragonphy_top with mdll instantiated and push it today

sgherbst:
@sjkim85 thanks, looking forward to the mdll update.  keep in mind that you'll need to add a stub or QTM for the mdll and possibly add constraints.  I'll add a wiki page on how to run the synthesis test to make this easier.

PR #86, opened 2020-05-11
Analog core2

sjkim85:
1) X-coupled S2D buffer is added for main sampler to match skew between CLK/CLKB
2) Unused ports are deleted from acore and jtag 
3) pnr scripts for butterphy_top is added (it's not going to be compatible with current flow)


sgherbst:
@sjkim85 looks like the issue is that signals ``disable_ibuf_test0`` and ``disable_ibuf_test1`` are referenced in the analog core debug interface, but not declared in the markdown file (it may need to be wired up in ``jtag`` and ``acore_debug_intf``.
https://github.com/StanfordVLSI/dragonphy2/blob/9a048795f08b231743916667922480b0aeeaa0d5/vlog/new_chip_src/top/dragonphy_top.sv#L75-L87

sgherbst:
@zamyers and @sjkim85 I can take a look at this so you can focus on PnR.  Since we have three open PRs, should we plan on the merge order being #86 (this one), then #90, then #91?

zamyers:
Hi Steven, I fixed the errors by removing the registers that Sung Jin no longer needs and the buffers that he no longer needs

So it passes the test and test_emu. However it fails Mflowgen tests

https://buildkite.com/stanford-aha/dragonphy2/builds/213#eeedbdcc-2b2c-4026-8b2a-4f93f275edc4

-Zach

Issue #88, opened 2020-05-13
JTAG Hello Test Correction


zamyers:
Currently uses a statically assigned JTAG ID value. The added ability to set the JTAG ID will cause this test to fail.

sgherbst:
Will be resolved once PR #91 is merged.

Issue #58, opened 2020-04-20
Second-pass PRBS checker features


sgherbst:
PR StanfordVLSI/dragonphy#78 contains the first pass of a PRBS checker, but more features are needed to improve testing capabilities and make debug easier.  I'm recording these features here in rough priority order:
1. Add ``FREEZE`` command for PRBS checker so that the number of total and correct bits are frozen while being read back over JTAG (essential feature).
2. Add mode for manually adjusting the fine and coarse delay of the PRBS checker.  This is important since it would allow us to do manual alignment, at least for relatively short PRBS sequences.
3. Make the ``ALIGN`` mode more robust to errors in the RX data stream.  The current implementation assumes that errors happen only rarely, but we should be prepared for more frequent errors (particularly when we are taking data directly from the ADC)
4. Add capability to single-step the PRBS generators and read back the PRBS generator output.
5. Add capability to produce simpler test patterns (e.g., square wave) to ease initial bringup.
6. Add more comprehensive gating to shut off the power consumption of the PRBS checker when not in use.
7. Support multiple LFSR lengths and equations (currently PRBS-7 is used, but it would be nice to support longer LFSRs too)
8. Add bit-twiddling escape hatches (e.g., invert RX data going into checker, invert PRBS pattern, reverse RX data)

sgherbst:
These changes are implemented in PR #90 .  Will close this once the PR is merged.

Issue #55, opened 2020-04-16
Make the JTAG ID include the commit hash and git clean/dirty status bit


sgherbst:
This is dependent on a JusTAG task: https://github.com/StanfordVLSI/JusTAG/issues/4
Once the JusTAG feature is ready, we can add the JTAG ID when ``make.py`` is run to build the chip.  (The current git commit hash can be retrieved with ``git rev-parse HEAD``.)

We have 31 bits available in the JTAG ID, and I suggest that we use the 28 most significant bits for the git commit hash, since it is customary to reference git commits using 7 nibbles.  I suggest we use the three remaining bits to represent additional information about the repository state.  For now, the main thing that comes to mind is the clean/dirty status of the repository to indicate whether the user has uncommitted changes (that can be automatically determined by checking if ``git status --porcelain`` is a non-empty string)

We can then encode this information in the ``version``, ``part``, and ``man_num`` variables:
* ``version[3:0] = (git_hash >> 24) & 0xf``
* ``part[15:0] = (git_hash >> 8) & 0xffff``
* ``man_num[10:0] = {(git_hash & 0xff), 1'b0, 1'b0, git_is_clean}``

For example, suppose the output of ``git rev-parse HEAD`` is
```shell
1abc49b8ec66bad372e5728d2a2f42ab1b1264fa
```
``git_hash`` would then be the most significant 7 nibbles (``1abc49b``).  Further assuming that the git repository is clean (``git_is_clean = 1'b1``), we have:
* ``version[3:0] = 4'h1 = 4'b0001``
* ``part[15:0] = 16'habc4 = 16'b1010101111000100``
* ``man_num[10:0] = {8'h9b, 1'b0, 1'b0, 1'b1} = 11'b10011011001``

The full JTAG ID (including the LSB that is always set to "1" by the TAP) would then be
```verilog
jtag_id = 32'b00011010101111000100100110110011 = 32'h1ABC49B3
```

The nice thing about this format is that we can immediately read off the git commit has using the most significant 7 nibbles, while the git status is encoded in the least significant nibble ("3" means clean, "1" means dirty).  We even have two more bits left over for other purposes if we can think of something :-)

sgherbst:
This change is implemented in PR #91.  Will close once the PR is merged.

Issue #32, opened 2020-04-08
Update handling of Vcal net


sgherbst:
In ButterPHY, we used a custom nettype for modeling the Vcal net, which can be driven externally or internally by a bias generator.  In DragonPHY, we're trying to support Xcelium and Vivado for most simulations, so this feature had to be temporarily removed (Vivado does not yet support a custom ``nettype``).  In most cases, this level of modeling detail is not needed, so it would be fine to have a "simple modeling" mode that supports Vivado and Xcelium, along with a "detailed modeling mode" that supports Xcelium only.  But we would need to do this in a way that doesn't require keeping two sets of code (since that's not maintainable).

sgherbst:
This is fixed in PR #90 and will be closed once that PR is merged.

PR #90, opened 2020-05-13
Update PRBS, Vcal, mflowgen, JTAG ID

sgherbst:
This PR makes four distinct updates:
* Pulls in recent updates from the **mflowgen** branch (custom-geom, custom-init, custom-power steps, using defaults for custom-init and custom-power for now -- @zamyers is working to update those scripts)
* Fixes the issue about the Vcal net (#32) so that it is now an inout net as expected
* Resolves the issue about PRBS updates (#58) so that it is more robust and flexible
* Resolves issues related to setting the JTAG ID to the git commit hash -- #55 (implementation) and #88 (testing the feature update)


zamyers:
LGTM

PR #91, opened 2020-05-13
Set JTAG ID using the git hash

sgherbst:
This PR resolves both #55 (implementation) and #88 (test update).  Once PR #90 is merged I'll sync this PR with those changes and request a reviewer.


sgherbst:
I'll add these changes to PR #90 so there is just one PR that has to be reviewed.  Closing this PR as a result.

PR #96, opened 2020-05-15
Analog core5

sjkim85:
V2T_clock_gen.sv  is modified to fix the en_sync sequence issue.
Now, data sequence at adc output looks good but it still looks wired after dcore.
I want @standanley to look at this

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,8 @@ pnr/data
 pnr/gate_size_test
 flow/*/outputs
 designs/*/*/temp/
+*.mco
+*.tar.gz
 
 #########################
 # other files to ignore #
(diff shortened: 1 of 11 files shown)


sjkim85:
@standanley , Can merge this to master?

sjkim85:
@standanley , Can you merge this to master?

standanley:
I just added a commit to change the en_sync chain so the signals should be in order. I am surprised to see that the AC test was passing because it looks like somebody already got rid of the magic re-ordering with the old en_sync chain. Was there a different change made that affected the ordering of ADC clocks?

sjkim85:
@standanley I've change the index ordering backed to normal order to test the fixed version in this PR

standanley:
I made a change to the default values for the config register codes for the retimer. I think before there was a problem with different ADCs having different latency. For me, this fixes issues with top_i.idcore.adcout_retimed when I look at the waveforms.

It does not affect the AC test because that pulls data from before the retimer.

sjkim85:
Thanks for your work.
As you know, we have to run AC test with final output of dcore (input to SRAM) eventually.

zamyers:
LGTM

PR #103, opened 2020-06-02
Merge emulation and simulation views

sgherbst:
This PR unifies the DragonPHY implementation used for simulation and emulation, removing a bunch of duplicated files and boilerplate code in the process.

## Details
1. There are no longer folders ``new_chip_src``, ``new_pack``, etc. -- the emulator is built from the same chip design.
2. You can now pass the command-line argument ``--dump_waveforms`` when running CPU simulations via pytest, rather than manually changing the code of the test in question. 
3. It is no longer necessary to use ``@pytest.mark.parameterize`` to mark regression tests; the folder path is used to determine which tests should be run on FPGA vs. CPU.
4. Other boilerplate code has been refactored into ``DragonTester`` to simplify the creation of new tests.
5. The hierarchy within the ``tests`` directory has been restructured so that we have ``cpu_block_tests``, ``cpu_system_tests``, ``fpga_system_tests``, ``fpga_block_tests``, and ``other_tests``.  This make it a bit more manageable to survey the test suite.
6. More BuildKite artifacts are added for easier debugging of failed regression runs (e.g., synthesis and PnR logs)
7. The emulation itself is fairly primitive at the moment; it uses bit-banging over the VIO interface to read out the JTAG ID of DragonPHY.  Future PRs will re-integrate generated analog/mixed-signal models and introduce an on-FPGA firmware layer for better performance (which is prototyped [here](https://github.com/sgherbst/jtag_test))


inc/fpga/iotype.sv
```diff
@@ -0,0 +1,10 @@
+`ifndef __IOTYPE_SV__
+`define __IOTYPE_SV__
+
+	`define pwl_t wire logic
```
standanley, line 4: For FPGA, why are pwl and real numbers just "wire logic"?

sgherbst: Good question -- this PR just gets the emulation of some digital blocks in the design up and running.  A future PR will integrate the synthesizable analog/mixed-signal models.

standanley: After reading the rest of the pull request, it looks like the models in the fpga folder are just dummy anyway

PR #109, opened 2020-07-23
Add runtime-updatable step response functions

sgherbst:
## Summary
This PR leverages new features in ``msdsl`` and ``svreal`` to model channel dynamics in a way that can be updated at runtime, without requiring the bitstream to be rebuilt.  The feature has been added to both high-level and low-level emulator architectures.  Emulator throughput was not affected by this change, but it was interesting to note that BRAM utilization is up while LUT utilization is down.  This suggests that Vivado synthesis had been using LUTs to implement some synchronous ROMs in the previous implementations of channel dynamics.

Here's an example of what updating the step response function at runtime looks like from a user perspective:
```python
placeholder = PlaceholderFunction(...)  # defines fixed-point formatting of piecewise-polynomial coefficients
chan_func = ...  # can use a regular Python function here
coeffs_bin = placeholder.get_coeffs_bin_fmt(chan_func)
coeff_tuples = list(zip(*coeffs_bin))
update_chan(coeff_tuples)  # transmits coefficients to FPGA over USB-UART
```

## Details
* As usual, most of the changes in the PR are in the testing infrastructure; the updates to the models themselves (``analog_slice`` and ``chan_core``) are about 25 lines each.
* The updatable functions can be configured via ``config/fpga/analog_slice_cfg.yml`` (high-level architecture) and ``config/fpga/chan.yml`` (low-level architecture)
* New ``pytest`` options ``--chan_tau`` and ``--chan_delay`` allow the user to conveniently specify the time constant and delay for the channel, assuming an exponential step response.  However, the user can provide their own function if an exponential step response is not sufficient, without having to rebuild the bitstream.

### DIFF
--- a/config/fpga/analog_slice_cfg.yml
+++ b/config/fpga/analog_slice_cfg.yml
@@ -1,6 +1,9 @@
 dt: 0.1e-6
 func_order: 1
 func_numel: 512
+func_domain: [0.0, 3.19375e-9]
+func_widths: [18, 18]
+func_exps: [-16, -16]
 chunk_width: 8
 num_chunks: 4
 slices_per_bank: 4
--- a/config/fpga/chan.yml
+++ b/config/fpga/chan.yml
@@ -1,4 +1,7 @@
 dt: 0.1e-6
 func_order: 1
 func_numel: 512
+func_domain: [0.0, 3.19375e-9]
+func_widths: [18, 18]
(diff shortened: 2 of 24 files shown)


sgherbst:
Switched reviewer to @zamyers since he will be running some tests on the emulator soon & may want to change the channel dynamics.

PR #113, opened 2020-09-06
Get better BER agreement between simulation and emulation

sgherbst:
## Summary

The main goal of this PR is to get simulation and emulation BER estimates to match better.  With the updates included here, the estimates are now match to within 7.5% or better at BER levels ranging from 1e-2 to 1e-4.  For BERs between 1e-5 and 1e-6, I have looked at the low-level and high-level emulators and found that they match to within the same tolerance. 

This PR is large mainly because it includes a lot a experimental code and results in the `experiments` folder.  Might be interesting to think about if there is a better home for that code.  I do like keeping experiments under revision control, though, because it makes it easier to replicate results.

## Details

There are several key updates to improve BER matching, both to the emulator implementations and the simulation baseline itself:
* Emulation
  * Switch from using an LFSR for random integer generation to using a Linear Congruential Generator (LCG); this is what the Verilog spec calls for.  While a Mersenne Twister would be "more random", it produces quite different results.  For now, it's simply left up to the user whether they want to use an LCG or Mersenne.
  * Use a pseudo-logarithmic compression in implementing the inverse CDF of a Gaussian.  This ends up focusing more PWL segments at parts of the function near zero, which are extremely sharp and nonlinear.
  * For the low-level emulator, fix the PI model so that it can handle delays that are slightly greater than one period (due to jitter).
  * For the high-level emulator, implement two details that are automatically covered by the low-level emulator and simulation baseline:
    1. Clamp jittery PI delays to be nonnegative
    2. Fix handling of PI delays that end up going beyond the `clk_adc` sampling point.
  * For both emulators, make the PRBS generator use the same equation as the mLingua PRBS21 module.  For the high-level case, this requires seeding the parallel PRBS generators in a certain way.
  * For both emulators, fix the TX and RX Vref levels to match simulation.
* Simulation
  * Fix two bugs in the baseline ADC model: noise was being added to the magnitude, rather than to the signal, and the ADC was rounding, instead of taking the floor.
  * Use tighter `etol` value for mLingua
  * Use a fast bit -> PWL transition in the TX driver so that the channel response is dominated by the first order pole.
  * Make the 8 GHz DragonPHY input clock derived from the 16 GHz TX clock to avoid simulator drift

More details follow as comments in the code.  It may be more convenient to view them by clicking on the "Files changed" tab, so that you can see the comments in context.


config/fpga/analog_slice_cfg.yml
```diff
@@ -1,16 +1,16 @@
 dt: 0.1e-6
 func_order: 1
 func_numel: 512
-func_domain: [0.0, 3.19375e-9]
+func_domain: [0.0, 1.0e-9]
```
sgherbst, line 4: tightened domain for step response to improve accuracy (given a fixed number of PWL segments).  this is still plenty for a time constant of 100ps.

config/fpga/analog_slice_cfg.yml
```diff
@@ -1,16 +1,16 @@
...
 func_widths: [18, 18]
 func_exps: [-16, -16]
 chunk_width: 8
 num_chunks: 4
 slices_per_bank: 4
 num_banks: 4
 pi_ctl_width: 9
-vref_rx: 0.4
-vref_tx: 0.4
+vref_rx: 0.3
```
sgherbst, line 12: In the simulation baseline, voltages vary from 0.1 to 0.4V, hence the change.  This was particularly an issue for ADC input-referred noise, because larger signals are less impacted by the same noise addition

config/fpga/chan.yml
```diff
@@ -1,7 +1,7 @@
 dt: 0.1e-6
 func_order: 1
 func_numel: 512
-func_domain: [0.0, 3.19375e-9]
+func_domain: [0.0, 1.0e-9]
```
sgherbst, line 4: Same comment as before; wanted to improve the accuracy of the step response.  Not sure if this was strictly necessary, though, since I ended up identifying other more significant causes of BER discrepancies.

config/fpga/clk_delay.yml
```diff
@@ -1,3 +1,4 @@
 dt: 0.1e-6
 n_bits: 9
 t_per: 250.0e-12
+depth: 3
```
sgherbst, line 4: In the low level emulation model, we can now handle PI delays that go from 0-1.5 periods of the 4 GHz clock (3 edges, not 2 as before)

dragonphy/adapt_fir.py
```diff
@@ -12,12 +12,14 @@ def __init__(self,  filename=None, **system_values):
...
         # create channel model
+        chan_resp_depth = 512
+        chan_samp_step = 1e-9/(chan_resp_depth-1)
         kwargs = dict(
             channel_type='exponent',
-            tau=25e-12,
-            t_delay=31.25e-12,
-            sampl_rate=160e9,
-            resp_depth=512
+            tau=100e-12,
```
sgherbst, line 19: using a more interesting value of tau -- using the FFE is required in order to get a reasonable BER with this value

dragonphy/fpga_models/analog_slice.py
```diff
@@ -29,8 +30,8 @@ def __init__(self, filename=None, **system_values):
                              real_type=get_dragonphy_real_type())
 
         # Random number generator seeds (defaults generated with random.org)
-        m.add_digital_param('jitter_seed', width=32, default=46428)
-        m.add_digital_param('noise_seed', width=32, default=8518)
+        m.add_digital_input('jitter_seed', width=32)
```
sgherbst, line 33: random seeds are now generally wired as signals, rather than as parameters.  i'm not sure if this is necessary, but i have seen bugs in Vivado's handling of parameter arrays in the past

dragonphy/fpga_models/analog_slice.py
```diff
@@ -108,25 +109,63 @@ def __init__(self, filename=None, **system_values):
...
-            t_samp = m.bind_name('t_samp', t_samp_pre + t_samp_jitter)
+
+            # add jitter to the delay amount (which might possibly yield a negative value)
+            delay_amt_noisy = m.bind_name('delay_amt_noisy', delay_amt_pre + delay_amt_jitter)
+
+            # make the delay amount non-negative
+            delay_amt = m.bind_name('delay_amt', if_(delay_amt_noisy >= 0.0, delay_amt_noisy, 0.0))
         else:
-            t_samp = t_samp_pre
+            delay_amt = delay_amt_pre
```
sgherbst, line 136: what follows is a bunch of code to handle scenarios where the sampling point occurs after the clk_adc edge.  for example, when the last sample was "early" and this sample is "late", we have to keep the output data constant and pass the sampling time onto the next clk_adc cycle.  this actually makes a significant difference at low BER

dragonphy/fpga_models/analog_slice.py
```diff
@@ -180,16 +219,23 @@ def __init__(self, filename=None, **system_values):
...
                 clk=m.clk,
                 rst=m.rst,
+                ce=m.write_output,
                 lfsr_init=m.noise_seed
             )
             sample_value = m.bind_name('sample_value', sample_value_pre + sample_noise)
         else:
             sample_value = sample_value_pre
 
+        # there is a special case in which the output should not be updated:
```
sgherbst, line 229: as mentioned before, there is a special case in which the ADC output is not updated and holds the same value from the last clk_adc cycle.  based on looking at logic analyzer waveforms, this really does happen from time to time

dragonphy/fpga_models/clk_delay_core.py
```diff
@@ -59,57 +57,116 @@ def __init__(self, filename=None, **system_values):
...
-            real_range_hint=m.emu_dt.format_.range_,
-            width=m.emu_dt.format_.width,
-            exponent=m.emu_dt.format_.exponent
-        )
-        m.set_this_cycle(m.dt_req, dt_req_array)
-
-        # set the output value to "1" if the rising edge DT request has been granted (if it's valid),
-        # set the output value to "0" if the falling edge DT request has been granted (if it's valid),
-        # otherwise k
+        m.bind_name('clk_edge', m.clk_i_val ^ m.clk_i_val_prev)
```
sgherbst, line 60: this is an implementation of a general clock signal delay that can keep a memory of the past "n" clock edges.  the main limitation is that the order of output edges has to match the order of input edges, but that is a realistic constraint

dragonphy/tester.py
```diff
@@ -24,9 +24,12 @@ def __init__(self, top_module='test', ext_test_bench=True, disp_type='realtime',
 
 		# adjust defines
 		defines['DAVE_TIMEUNIT'] = '1fs'
-		defines['NCVLOG'] = None
+		if simulator == 'vcs':
```
sgherbst, line 27: i added some experimental support for VCS.  i'm not yet convinced it is yielding accurate results, though.  i think we would need to go back to mLingua and check the Makefiles to make sure that I'm using all of the required flags here

experiments/cpu_emu_comparison/phase_interpolator.sv
```diff
@@ -61,7 +61,7 @@ module phase_interpolator #(
         delay_s = ((1.0*ctl)/(2.0**(Nbit)))*(250.0e-12);
 
         // add jitter
-        delay_s += (`JITTER_RMS)*($dist_normal(seed, 0, 1000)/1000.0);
+        delay_s += (`JITTER_RMS)*($dist_normal(seed, 0, 10000000)/10000000.0);
```
sgherbst, line 64: I experimented with values from 100 to 1,000,000,000 and found that anything between 1,000 and 100,000,000 yields similar results.  this particular setting gives a resolution of 0.1uV, but that is not necessary for accurate results

experiments/cpu_emu_comparison/stochastic_adc_PR.sv
```diff
@@ -122,21 +123,29 @@ module stochastic_adc_PR #(
...
             out_sgn = 1;
-            samp = samp_p - samp_n;
         end else begin
             out_sgn = 0;
-            samp = samp_n - samp_p;
+            samp = -1.0*samp;
         end
 
-        // add noise
-        samp += (`NOISE_RMS)*($dist_normal(seed, 0, 1000)/1000.0);
```
sgherbst, line 136: this was a bug -- i should not have taken the absolute value before adding noise

experiments/cpu_emu_comparison/stochastic_adc_PR.sv
```diff
@@ -122,21 +123,29 @@ module stochastic_adc_PR #(
...
-
         // determine output magnitude
-        out_mag = ((1.0*samp) / (0.3)) * ((2.0**(Nadc-1.0))-1.0);
```
sgherbst, line 139: this implicitly rounded to get out_mag, which didn't match the emulator and probably is not what we wanted

experiments/cpu_emu_comparison/test.sv
```diff
@@ -107,16 +184,19 @@ module test;
...
 
-	clock #(
-		.freq(8.0e9), // This depends on the frequency divider in the ACORE's input buffer
-		.duty(0.5),
-		.td(0)
-	) iEXTCLK (
-		.ckout(ext_clkp),
-		.ckoutb(ext_clkn)
-	);
+    initial begin
```
sgherbst, line 191: deriving the 8GHz clock from the 16 GHz clock so that they stay locked, even during long simulations (they can drift due to numerical issues otherwise)

experiments/cpu_emu_comparison/test.sv
```diff
@@ -293,8 +376,8 @@ module test;
 
 		// Wait for MM_CDR to lock
 		$display("Waiting for MM_CDR to lock...");
-		for (loop_var=0; loop_var<2; loop_var=loop_var+1) begin
-		    $display("Interval %0d/2", loop_var);
+		for (loop_var=0; loop_var<100; loop_var=loop_var+1) begin
```
sgherbst, line 379: startup can be slower when there is noise, so i increased the waiting period

experiments/emu_vs_emu_macro/experiment.py
```diff
@@ -0,0 +1,353 @@
+import time
```
sgherbst, line 1: sorry, this is basically a copy of tests/fpga_system_tests/emu/test_emu.py.  code in the `experiments` folder is rougher than in the rest of the repository.  might be interesting to think about whether there is a better way to manage it

experiments/gaussian/gen.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
```
sgherbst, line 1: this experiment was used to test out the CDF of the msdsl-based Gaussian noise generator.  i found that it matched well to +/- 6 sigma.  of course, there is more to random number generation than just matching the CDF, but this was already a big improvement as compared to the approach before using logarithmic compression

experiments/lines_of_code/hl.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
sgherbst, line 1: the code in this directory was just used to count lines of code, to gather productivity numbers

setup.py
```diff
@@ -12,8 +12,8 @@
 
 requires_list = [
     # anasymod ecosystem
-    'svreal==0.2.5',
-    'msdsl==0.3.2',
+    'svreal==0.2.6',
```
sgherbst, line 15: svreal and msdsl got noise-related updates

tests/fpga_block_tests/adc_model/test_adc_model.py
```diff
@@ -25,14 +29,14 @@ def model(in_, n, vref):
 
     # determine magnitude
     in_abs = abs(in_)
-    mag_real = (abs(in_abs) / vref) * ((2**(n-1)) - 1)
+    mag_real = (abs(in_abs) / CFG['vref']) * ((2**(CFG['n']-1)) - 1)
```
sgherbst, line 32: there was not a single source of truth here, so i'm trying to fix that with this change

tests/fpga_system_tests/emu/main.c
```diff
@@ -25,6 +25,9 @@ void do_init() {
    set_jitter_rms_int(0);
    set_noise_rms_int(0);
 
+   // prbs control
+   set_prbs_eqn(0x100002);
```
sgherbst, line 29: the PRBS equation can now be changed at runtime.  some of the shorter PRBS sequences, like PRBS-7, have a very clearly apparent repetition that distorts the BER

tests/fpga_system_tests/emu/test_emu.py
```diff
@@ -364,13 +381,15 @@ def update_chan(coeff_tuples, offset=0):
...
+    write_tc_reg('retimer_mux_ctrl_1', 0b1111111111111111)
+    write_tc_reg('retimer_mux_ctrl_2', 0b1111111111111111)
+    #write_tc_reg('retimer_mux_ctrl_1', 0b0000111111110000)
+    #write_tc_reg('retimer_mux_ctrl_2', 0b1111000000000000)
 
     # Configure the CDR
     print('Configuring the CDR...')
     write_tc_reg('cdr_rstb', 0)
-    write_tc_reg('Kp', 15)
+    write_tc_reg('Kp', 18)
```
sgherbst, line 392: this didn't appear to make a difference to BER, but it's good to be consistent because I was using 18 everywhere else

tests/fpga_system_tests/emu_macro/simctrl.pre.yaml
```diff
@@ -82,6 +89,191 @@ digital_probes:
     width: 10
     signed: 1
 
+  # handling of rare cases in analog_slice
+  t_samp_idx_0:
+    abspath: 'tb_i.top_i.iacore.genblk1[0].analog_slice_i.t_samp_idx'
```
sgherbst, line 94: these probes could likely be commented out, but it's interesting to see when "unusual" events happen that mess up the sequencing of the ADC samples, since that contributes to BER

tests/fpga_system_tests/emu_macro/tb.sv
```diff
@@ -97,8 +119,8 @@ module tb;
                 .clk(emu_clk),
                 .rst(emu_rst),
                 .cke(prbs_cke),
-                .init_val(i+1),
```
sgherbst, line 100: this was not the right way to seed the PRBS generators to match the output of a single generator

vlog/tb/diff_channel.sv
```diff
@@ -4,6 +4,7 @@
 
 module diff_channel #(
 	parameter channel_type = "simple",
+	parameter real etol = 0.001,  // error tolerance of PWL approximation
```
sgherbst, line 7: etol does make a small difference for BER estimation, particularly for jitter as it can change the slope of the RX input

standanley (approved):
Hi Steven, the PR looks good! I looked a litter closer at the code to do with edges falling outside the cycle they typically fall in, and it makes sense to me. I don't know if I have a good enough understanding of the whole environment to say whether there is a better (i.e. less special case code) way to hand this, but I believe that the current code works and is clear.

PR #114, opened 2020-09-21
Add ZCU106 support

sgherbst:
## Summary
This small PR adds support for running emulations on the ZCU106 board.  We should hold off on approving it until it is confirmed to be working for @zamyers.

## Details
1. ``setup.py`` is updated to use the most recent version of **anasymod**, which now has support for ZCU106.
2. A memory corruption bug was fixed in the ``main.c`` firmware files used for the low- and high-level emulators.  I'm not sure why this only showed up for ZCU106, not ZC706 or other boards, but it appears to be related to how null character termination was being done and whether ``unsigned`` vs. ``unsigned long`` was used as the format in ``sscanf``.


zamyers (approved):
LGTM

Issue #115, opened 2020-09-21
Add long-duration histogram capability


sgherbst:
This feature would be used to explore ADC noise.  At the moment we can only capture batches of ~1000 points at a time, but to study the tails of the distribution, we need to include orders of magnitude more samples.

PR #116, opened 2020-09-23
Add histogram measurement feature

sgherbst:
## Summary

This feature resolves issue #115 by adding a histogram block to DragonPHY.  This is intended to be used to study ADC noise, examine the bimodal distribution of data at the FFE output, and possibly to construct eye diagrams.  Users can specify the histogram input to be: (1) any ADC output (including replicas), (2) any FFE output, and (3) an on-chip data generator.  This last option can produce data with various distributions, for histogram.

The implementation includes two small SRAMs (each 256x64).  I thought about instantiating the exact same SRAM that is already used for waveform capture (1024x144), but since that SRAM is 9x larger than what is needed for the histogram, I figured it would be worth it to generate another SRAM.  I updated the **mflowgen** steps accordingly through synthesis, but affected PnR steps such as floorplanning will have to be updated by someone with process access (that person should also take a look at the new `mc-gen-sram-small` step to make sure it's right).

Note: this PR is independent from PR #114 and they can be merged in either order.

## Details
1. The histogram is implemented using SRAM, where each memory address corresponds to one possible value of the ADC output.  Since the SRAM is single-port, incrementing the value at a particular memory address is a two-cycle process.  Hence, two SRAMs are used in an alternating fashion to log all of the incoming data.  When the user reads the histogram value at a particular address, the outputs of both SRAMs at that address are summed.
2. The histogram feature is used by setting `hist_mode` to `RESET`, then `CLEAR` for a short period of time to clear the SRAMs (1us is plenty), and finally `RUN` to capture data.  When the user is ready to read the histogram, they change `hist_mode` to `FREEZE`, which stops the histogram from updating.  The `hist_count` at each address can be read out, as well as the `hist_total` of all addresses.  These values are 64-bit, which would allow for hundreds of years of operation without overflowing (for the true aficionados of distribution tails :-)).  Each 64-bit value is split into 32-bit `upper` and `lower` JTAG registers.
3. Histogram addresses should be interpreted in twos-complement when working with ADC or FFE data, which is signed.  So the address `8'b11111111` (255) refers to an ADC output of `-1`.
4. The source of histogram data can be selected using the `hist_source` and `hist_src_idx` JTAG registers.  For `hist_source`, `0` means "ADC", `1` means FFE, and `3` means the data generator used for BIST.  (If we have any outputs from MLSD that could be logged using a histogram, those could be mapped to `2`).  The `hist_src_idx` setting only applies when the source is "ADC" or "FFE", in which case it references to the ADC slice / FFE channel.  For the ADC, indices `16` and `17` are replica ADCs.  It might be interesting to monitor the noise histogram of a replica with and without the other ADCs running to study power supply noise.
5. The BIST data generator is mainly controlled by JTAG registers `data_gen_mode`, `data_gen_in_0`, and `data_gen_in_1`.  When `data_gen_mode` is `UNIFORM`, the data generator produces a uniform distribution between 0-255.  This data stream is not pseudorandom; the values are simply produced sequentially.  Another mode is `CONSTANT`, where the data generator produces a steady value of `data_gen_in_0`.  `INCL` is a more general version of `UNIFORM`, producing a uniform distribution between `data_gen_in_0` and `data_gen_in_1`.  `EXCL` is the opposite of `INCL`, producing values outside the interval between `data_gen_in_0` and `data_gen_in_1`.  Finally, `ALT` alternates between `data_gen_in_0` and `data_gen_in_1` on every cycle of `clk_adc`.
6. I added new **mflowgen** steps to generate the smaller SRAM: `openram-gen-sram-small` and `mc-gen-sram-small`.  The `openram-gen-sram-small` step is exercised through our regression testing flow, but I do not have a way to test `mc-gen-sram-small` myself, since I do not have process access.
7. I am using the `System` clock domain for all of the JTAG registers (input and output) related to the histogram.  I am assuming this synchronizes I/O to `clk_adc` -- it is particularly important that `hist_mode` is synchronized, since otherwise the transitions in the histogram state machine could be messed up.  (We should confirm what synchronization is being done -- maybe @zamyers would know?)


standanley (approved):
This PR looks good! I have a couple notes but nothing that blocks the merge.

There's a TODO in the digital core that asks about using MLSD output as histogram data. That seems like a good tool to have and is probably relatively low-effort to implement; we should remember to bring it up at the next circuits meeting.

In the CPU models and FPGA models it looks like sram.sv and sram_small.sv are the same except for the default sizes. It might make sense to implement sram_small as an instantiation of sram with different parameters, although it would make the hierarchy a little more confusing.

PR #122, opened 2020-10-01
PFD offset calibration loop updates

sgherbst:
## Summary
This PR is an update to the PFD offset calibration feature, including several new block- and system-level tests.  The main implementation change has to do with the `update` signal used to reset averages and histograms: previously a generated clock signal was involved, but this has now been replaced with a clock enable signal.  The result is a simpler synthesis flow, both for the ASIC and for FPGA emulation.

## Details
* Added a system-level test of PFD offset calibration (`pfd_offset`).  This is basically a DC sweep, but the PFD offset is determined by the calibration loop, rather than set externally.
* Added block-level tests of modules related to PFD offset calibration (`adc_unfolding`, `avg_pulse`, `calib_hist`, `sample_avg`).  These tests account for most of the new code in this PR.
* Reviewed bitwidths in the calibration circuitry and increased a few that were at risk of overflow.
* The JTAG register values of `adcout_hist_center` and `adcout_hist_side` are now sampled only at the end of each `update` cycle, so that users don't end up reading a random intermediate value.
* Added a few JTAG registers for debugging:
  * `pfd_cal_flip_feedback`: A "chicken" bit that can be used to flip the polarity of the PFD offset calibration feedback loop.  (Disabled by default)
  * `en_pfd_cal_ext_ave`: Disabled by default, but when set to `1`, the averaging block in the offset calibration is bypassed and replaced with a user-specified value, `pfd_cal_ext_ave`.  In addition to being useful for debugging the offset calibration, it could also be used to read out histogram values one-by-one (i.e., set `Nbin` to `0`, then set `pfd_cal_ext_ave` to the desired ADC code).


sjkim85 (approved):
For me, all changes make sense.
I like flipping polarity by a chicken bit and bypassing avg block for testing.

PR #125, opened 2020-10-07
analog_core updates

sjkim85:
1. phase_interpolator is updated 
  - nand based delay chain
  - "en_unit" is added
 2. stocahstic_adc_PR is updated
  - clk_async is removed
  - phase_monitor is removed
  - TDC clock selection is removed
  - PFD input mux is added
  - "sel_PDF_in" and "sign_PFD_clk_in" are added
3. analog_core is updated
  - not used ports are removed
4. jtag registers are updated


sgherbst:
Hey @sjkim85, I took a quick look at the error messages and noticed the following issues:
1. Since ``pfd_inp_meas``/``pfd_inn_meas`` were removed from the analog core debug interface, I think you want to set the right-hand side of these two assignments to ``0`` in ``digital_core``:
https://github.com/StanfordVLSI/dragonphy2/blob/2b71913c3d7c6279db99a17f0a895980ecff1ebf/vlog/chip_src/digital_core/digital_core.sv#L505-L506
2. Since you changed the analog debug interface, please update the QTM for ``analog_core`` accordingly.  Here's the path: ``dragonphy2/designs/dragonphy_top/qtm/analog_core.qtm.tcl``
3. Please take a look at the model of ``analog_core`` used in the high-level emulator and delete assignments to analog debug signals that were removed.  At a minimum, I think these lines need to be deleted:
https://github.com/StanfordVLSI/dragonphy2/blob/2b71913c3d7c6279db99a17f0a895980ecff1ebf/vlog/fpga_models/analog_core/analog_core.sv#L217-L219
https://github.com/StanfordVLSI/dragonphy2/blob/2b71913c3d7c6279db99a17f0a895980ecff1ebf/vlog/fpga_models/analog_core/analog_core.sv#L225-L227
https://github.com/StanfordVLSI/dragonphy2/blob/2b71913c3d7c6279db99a17f0a895980ecff1ebf/vlog/fpga_models/analog_core/analog_core.sv#L236-L237

sgherbst:
I reviewed the errors and saw a couple of issues:

1. ``test_mflowgen``
```text
Error: Can't find inout port 'adbg_intf_i.sel_PFD_in' on reference to 'analog_core' in 'dragonphy_top'. (LINK-1)
```
Looks like you didn't add analog_core QTM pins for all of the new registers.  I just did a quick scan, but I think ``sel_PFD_in``, ``sign_PFD_clk_in``, ``sel_PFD_in_rep``, and ``sign_PFD_clk_in_rep`` are missing from the QTM.

2. ``test_emu``
```text
ERROR: [VRFC 10-3180] cannot find port 'sign_PFD_clk_in' on this module [/var/lib/buildkite-agent/builds/fpga-verif-1-1/stanford-aha/dragonphy2/vlog/chip_src/analog_core/analog_core.sv:221]
ERROR: [VRFC 10-3180] cannot find port 'sel_PFD_in' on this module [/var/lib/buildkite-agent/builds/fpga-verif-1-1/stanford-aha/dragonphy2/vlog/chip_src/analog_core/analog_core.sv:220]
ERROR: [VRFC 10-3180] cannot find port 'sign_PFD_clk_in' on this module [/var/lib/buildkite-agent/builds/fpga-verif-1-1/stanford-aha/dragonphy2/vlog/chip_src/analog_core/analog_core.sv:258]
ERROR: [VRFC 10-3180] cannot find port 'sel_PFD_in' on this module [/var/lib/buildkite-agent/builds/fpga-verif-1-1/stanford-aha/dragonphy2/vlog/chip_src/analog_core/analog_core.sv:257]
```
Please add the input pins ``sign_PFD_clk_in`` and ``sel_PFD_in`` to the emulation model ``dragonphy2/vlog/fpga_models/analog_core/stochastic_adc_PR.sv``.  You don't have to implement their function, but they should at least be present in the port list, otherwise FPGA synthesis will fail.

3. This isn't related to an error, but please delete obsolete code rather than commenting it out.  We can always restore the old code from the git history, but it can be confusing to read/review code when a lot of it is commented out.  I would typically only leave in a commented block of code if there was a remark explaining it (i.e., "uncomment this code to create an interactive plot").  The main files I have in mind here are ``analog_core.qtm.tcl``, ``acore_debug_intf.sv``, ``analog_core.sv`` (FPGA & ASIC versions), ``stochastic_adc_PR.sv``, and ``jtag.sv``.

sjkim85:
- Input clock network(clk_in) of the phase_interpolator has been changed for this tapr-out. (initial commit)
  (two explicit gates are removed from clk_in network --> 30ps off from ADC's ideal sampling phase in cpu simulation)
- keep pi_ctl but add 30p delay to ext_clk in test bench (cpu_system_test/prbs,  cpu_system/tests/loopback_sram, cpu_system_tests/loopback_retimer)
- minimum v2t gain is adjusted from 240 to 210 in cpu_systems/pfd_offset 
- all cpu tests in local machine pass 
- bit overflow issue in jtag register (unq16) is still not solved (warning, not error)

sgherbst:
Great progress!  Looks like only the FPGA emulation is failing.  I suspect that may be due to the JTAG register overflow warning -- we don't have any CPU-based tests that thoroughly exercise the JTAG interface, but that is exercised more fully in emulation.  So emulation may be catching a real bug.

There is a quick test that you can do to look for JTAG build errors:
1. Checkout ``dragonphy/jtag.py`` from the ``small_fixes`` branch (not the whole commit, just that one file).  This fixes an issue where JTAG build errors weren't being properly reported:
```shell
> git fetch
> git checkout small_fixes -- dragonphy/jtag.py
```
2. Rebuild the JTAG registers
```shell 
> rm -rf build
> python make.py --view cpu
```
3. See if there are any exceptions, which can happen if you run out of registers.

If you're running out of registers, a quick fix is to consolidate a couple of one-bit registers into a single multi-bit register.  This is what I did in the ``small_fixes`` branch with the ``misc_ctrl_bits`` register.  If you don't see any new errors, then we probably have to dig further into the JTAG generator to figure out what's going on.

One unrelated question -- why does the V2T gain have to be adjusted in PFD offset?  Does this have to do with changes to the ADC?

sjkim85:
Steven,

I followed the instructions you guided but I could not find any error or exception.
(the bit overflow warning still exists)
Actually, if I compare "acore_intf.md" (between master and analog_core_update), the number of registers is decreased from 71 to 62.
So, I think it is not running out of register.
Do you have any other idea to test about the emulation fail?

-Sung-Jin

sgherbst:
Hi @sjkim85 -- thanks for checking that.  I still suspect it is related to the JTAG block, since if there are bitwidth issues in the register address decoding, that could mean that JTAG reads/writes aren't being executed properly.  @zamyers, do you have some thoughts on this, given that it doesn't look like Sung-Jin is running out of registers?

sgherbst (approved):
LGTM -- thanks @zamyers and @sjkim85 for putting in the effort to get the tests passing.

PR #127, opened 2020-10-10
The transmitter directory created

CansWang:
16 to 1 quarter-rate transmitter created


vlog/chip_src/tx_16t1/qr_4t1_mux_top.sv
```diff
@@ -0,0 +1,53 @@
...
+
+// 4 to 1 mux 
+//wire [1:0] sel;
+div_b2 div (.clkin(clk_IB), .clkout(ck_b2));
+// Combinational logic to generate the selection window from clk Q-QB-I-IB
+// assign sel =  
+
+always @(D1MQ or D1MI or D2MQB or D2MIB or clk_Q or clk_QB or clk_I or clk_IB) begin
+    assign data = (clk_Q && clk_I) ? D1MQ : ((clk_I && clk_QB) ? D1MI : ((clk_QB && clk_IB) ? D2MQB : D2MIB)); // Maybe problematic to write it in this way, need to check after synthesis
+end
```
sgherbst, line 49: If this is intended to be combination logic, please change to ``always @(*)`` and remove the ``assign`` keyword.

sgherbst:
Hi @CansWang -- thanks for submitting this PR!  Couple of comments:
1. I noticed that the tests aren't passing because it duplicates the definition of ``prbs_generator_syn.sv``.  In our system, each file should define exactly one module, whose name is given by the file name (e.g., ``prbs_generator_syn.sv`` defines exactly one module, called ``prbs_generator_syn``).  I think your code is already following that convention.  In addition, there should only be one definition of a module in each folder in ``vlog`` -- so ``chip_src`` can only define ``prbs_generator_syn`` once.  In this particular case, I think you can just remove that module from the ``tx_16t1`` folder.  Our dependency management system should pick up the location of that module automatically.
2. For the same reason, please remove the files ``ff_c`` and ``mux``, since they are already defined in ``analog_core/gates``.
3. The ``chip_src`` folder should only have code that is intended to go on the chip, so the ``tx_top_tb.sv`` file should be moved to one of the block- or system-level tests, or maybe to the ``vlog/tb`` folder.

CansWang:
> Hi @CansWang -- thanks for submitting this PR! Couple of comments:
> 
> 1. I noticed that the tests aren't passing because it duplicates the definition of `prbs_generator_syn.sv`.  In our system, each file should define exactly one module, whose name is given by the file name (e.g., `prbs_generator_syn.sv` defines exactly one module, called `prbs_generator_syn`).  I think your code is already following that convention.  In addition, there should only be one definition of a module in each folder in `vlog` -- so `chip_src` can only define `prbs_generator_syn` once.  In this particular case, I think you can just remove that module from the `tx_16t1` folder.  Our dependency management system should pick up the location of that module automatically.
> 2. For the same reason, please remove the files `ff_c` and `mux`, since they are already defined in `analog_core/gates`.
> 3. The `chip_src` folder should only have code that is intended to go on the chip, so the `tx_top_tb.sv` file should be moved to one of the block- or system-level tests, or maybe to the `vlog/tb` folder.

CansWang:
> Hi @CansWang -- thanks for submitting this PR! Couple of comments:
> 
> 1. I noticed that the tests aren't passing because it duplicates the definition of `prbs_generator_syn.sv`.  In our system, each file should define exactly one module, whose name is given by the file name (e.g., `prbs_generator_syn.sv` defines exactly one module, called `prbs_generator_syn`).  I think your code is already following that convention.  In addition, there should only be one definition of a module in each folder in `vlog` -- so `chip_src` can only define `prbs_generator_syn` once.  In this particular case, I think you can just remove that module from the `tx_16t1` folder.  Our dependency management system should pick up the location of that module automatically.

**Yes, I will remove them accordingly.
  Thanks for the reminder
**

> 2. For the same reason, please remove the files `ff_c` and `mux`, since they are already defined in `analog_core/gates`.
> 3. The `chip_src` folder should only have code that is intended to go on the chip, so the `tx_top_tb.sv` file should be moved to one of the block- or system-level tests, or maybe to the `vlog/tb` folder.

**I will reorganize the files and push it again**

vlog/chip_src/tx_16t1/qr_4t1_mux_top.sv
CansWang: This is a mux, I was trying to avoid `if` statement as mentioned in the previous meeting.

I will revise this part accordingly. Thx for the suggesitons

sgherbst: Thanks for keeping that in mind.  In that case you could simply delete the ``always`` and ``end``, leaving only the ``assign`` statement.  In that case you would need to change ``output reg data`` to ``output wire data``.  In fact, I think that is probably the cleanest solution.

sgherbst (approved):
Looks good, thanks!

PR #128, opened 2020-10-14
Transmitter

CansWang:
Add control bits and phase interpolator to the tx_top for top-level integration


vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
+    
+    input wire rst, // Global reset for Tx
+    input wire cke, // Gloabl lock gating for Tx
+
+    // Control and input of the phase interpolator
+    // input tx.rstb, // rst_bar for pi
+
+
+    // input tx.clk_in_pi,
+    input tx.clk_async, // Asynchronous clock for phase measurement
```
sgherbst, line 17: I suggest putting these debug signals into an interface, as we have done for the analog core.  You could start by copying ``dragonphy2/vlog/chip_src/analog_core/acore_debug_intf.sv`` to ``dragonphy2/vlog/chip_src/tx_16t1/tx_debug_intf.sv`` and then modifying the interface as needed.  I think the signals that you need are mostly the PI- and input divider-related signals from ``acore_debug_intf``, plus a few extras.

You'll notice that ``acore_debug_intf`` has two ``modports``, ``acore`` and ``dcore``.  I think you would rename the ``acore`` modport to ``tx``, so that in the ``tx_top`` I/O list, you could just write:
```systemverilog
module tx_top (
   ...
   tx_debug_intf.tx tx,
   ...
); 
```
instead of having to write out all of the ``tx.*`` signals individually.  Within the module definition for ``tx_top``, I don't think that you would have to change anything since you can refer to interface members using the same "dot" syntax.

This should make it easier to update the list of debug signals, and should also reduce the likelihood of typos.  It also fits in nicely with the way that we have set up JTAG debug registers (each interface has a corresponding Markdown file in ``dragonphy2/md``.

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
 module tx_top (
     input wire [15:0] din,
-    input wire clk_q,  // The clock inout must follow this order, (rising edge order) Q->I->QB->IB-Q
-    input wire clk_i,  // The clock is a quarter-rate clock with respect to output data rate
-    input wire clk_qb, // q, i, qb, ib spaced evenly within a clock cycle
-    input wire clk_ib,
-    output reg clk_prbsgen,  // Output clock for 16-bit prbs generator
-    output reg dout // Data output
+    input wire mdll_clk, // Clock from MDLL
+    input wire ext_clk, // Clock from external source
```
sgherbst, line 7: Since ``ext_clk`` also goes to ``analog_core``, please check with @sjkim85 that the additional loading by ``tx_top`` on this clock signal is OK.

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
+wire residue_buf1;
+wire residue_buf2;
+
 wire clk_halfrate;  // Input clock for 16 to 4 mux
+wire clk_q;  // The clock inout must follow this order, (rising edge order) Q->I->QB->IB-Q
+wire clk_i;  // The clock is a quarter-rate clock with respect to output data rate
+wire clk_qb; // q, i, qb, ib spaced evenly within a clock cycle
+wire clk_ib;
+
+wire [3:0] tx.clk_interp_slice; // Output from the phase interpolator
```
sgherbst, line 84: Please avoid using a "dot" in signal name declarations

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
+wire [3:0] tx.clk_interp_sw; //
+
+
+
+wire tx.clk_in_pi;
+wire tx.rstb;
+tx.rstb = ~ rst;
+
+//Global clock gating 
+assign clk_prbsgen_reg = cke ? 1b'0 : clk_prbsgen; //Output clock gating
```
sgherbst, line 94: Two notes here:
1. It is not allowed to ``assign`` to a ``reg`` -- you can only assign to a ``wire``.  I think you could solve this problem by changing ``clk_prbsgen_reg`` to a ``wire`` (and probably change the name, too, since it is not a ``reg``).  This might help to clear things up: https://www.verilogpro.com/verilog-reg-verilog-wire-systemverilog-logic/.
2. In general, I would advise against adding logic in the clock path like this, since it may be problematic for closing timing in synthesis, and also because it can introduce clock glitches.  I prefer to use ``cke`` in the sequential logic that uses the clock, e.g.
```verilog
always @(posedge clk) begin
    if (rst) begin
        q <= 0;  // reset output
    end else if (cke) begin
        q <= d;  // clock enabled
    end else begin
        q <= q;  // clock disabled; do nothing
    end
end
```
Automated clock gating in synthesis should recognize this pattern and insert clock gating circuit if it is advantageous from a power perspective.  I don't have a great reference for this, but you could look at Fig. 2 and Fig. 3 in this paper to get a better sense of what I'm talking about: https://ieeexplore-ieee-org.stanford.idm.oclc.org/stamp/stamp.jsp?tp=&arnumber=5733874

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
+
+wire tx.clk_in_pi;
+wire tx.rstb;
+tx.rstb = ~ rst;
+
+//Global clock gating 
+assign clk_prbsgen_reg = cke ? 1b'0 : clk_prbsgen; //Output clock gating 
+assign tx.ext_clk = cke ? 1b'0 : ext_clk;  // Input external clock gating
+assign tx.mdll_clk = cke ? 1b'0 : mdll_clk; // Input external clock gating
+assign residue_buf1 = 1 ? 1b'0 : residue_clk_halfrate; // This buf act as a buffer to filter any potential interference of the floating internal clock node.
```
sgherbst, line 97: Sorry, I don't quite follow what the purpose of these two lines is -- it look like ``residue_clk_halfrate`` and ``residue_clk_prbsgen`` are driven, not floating.

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
+    .clk_Q(tx.clk_interp_slice[0]),  // Quarter-rate clock input
+    .clk_QB(tx.clk_interp_slice[2]),
+    .clk_I(tx.clk_interp_slice[1]),
+    .clk_IB(tx.clk_interp_slice[3]),
+    .din(qr_data_p), // Quarter-rate data from half-rate 16 to 4 mux
+    .rst(rst),
+    .cke(cke),
     .ck_b2(clk_halfrate), // Divided quarter-rate clock for 16 to 4 mux
-    .data(dout) // Final data output
+    .data(dout_p) // Final data output
```
sgherbst, line 194: Just to check, I think ``dout_p`` and ``dout_n`` don't yet have the output drivers or termination added yet, right?

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: The differential signal is generated by calling the 16 to 4 and 4 to 1 mux once again with inverting inputs `~din`. The extra clock divider signal will be floating, just to ensure there are no uncertain connections, I put a gating on it.

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: No, I haven't instantiated the output buffer and termination on it yet.

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
+
+
+
+wire tx.clk_in_pi;
+wire tx.rstb;
+tx.rstb = ~ rst;
+
+//Global clock gating 
+assign clk_prbsgen_reg = cke ? 1b'0 : clk_prbsgen; //Output clock gating 
+assign tx.ext_clk = cke ? 1b'0 : ext_clk;  // Input external clock gating
```
sgherbst, line 95: I don't think gating circuitry is needed ``ext_clk`` and ``mdll_clk``, since they go directly to the ``input_divider``, which already handles clock selection and can be disabled.

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,218 @@
...
+                .max_sel_mux(tx.max_sel_mux[k])
+            );
+            assign tx.pi_out_meas[k] = (tx.sel_meas_pi[k] ?
+                                                 tx.clk_interp_slice[k] :
+                                                 tx.clk_interp_sw[k]) & tx.en_meas_pi[k];
+        end
+    endgenerate
+
+// Clock input divider
+	input_divider indiv (
```
sgherbst, line 149: Please mention to @sjkim85 that you want to use his ``input_divider`` block.  I think that he is already making some updates so that ``phase_interpolator`` can be used as a standalone block, so please check if similar updates are needed to use ``input_divider`` in the transmitter.

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: Okay, sure. I will dig into that and make corresponding changes.

> I suggest putting these debug signals into an interface, as we have done for the analog core. You could start by copying `dragonphy2/vlog/chip_src/analog_core/acore_debug_intf.sv` to `dragonphy2/vlog/chip_src/tx_16t1/tx_debug_intf.sv` and then modifying the interface as needed. I think the signals that you need are mostly the PI- and input divider-related signals from `acore_debug_intf`, plus a few extras.
> 
> You'll notice that `acore_debug_intf` has two `modports`, `acore` and `dcore`. I think you would rename the `acore` modport to `tx`, so that in the `tx_top` I/O list, you could just write:
> 
> ```systemverilog
> module tx_top (
>    ...
>    tx_debug_intf.tx tx,
>    ...
> ); 
> ```
> 
> instead of having to write out all of the `tx.*` signals individually. Within the module definition for `tx_top`, I don't think that you would have to change anything since you can refer to interface members using the same "dot" syntax.
> 
> This should make it easier to update the list of debug signals, and should also reduce the likelihood of typos. It also fits in nicely with the way that we have set up JTAG debug registers (each interface has a corresponding Markdown file in `dragonphy2/md`.

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: > Please avoid using a "dot" in signal name declarations

Noted. Will modify this.

vlog/chip_src/tx_16t1/tx_top.sv
sgherbst: Hmm, I usually think of a "floating" signal as being connected to one or more inputs, but zero outputs.  However, it looks like ``residue_clk_halfrate`` is driven by ``qr_mux_4t1_1.ck_b2`` and ``residue_clk_prbsgen`` is driven by ``hr_mux_16t4_1.clk_b2``.  In addition, ``residue_clk_halfrate`` is connected to the input ``hr_mux_16t4_1.clk_hr``.

For ``residue_clk_prbsgen`` at least, if you meant that it isn't connected to any input, you could use the ``()`` syntax to indicate that it is unused:
```verilog
hr_16t4_mux_top hr_mux_16t4_1 (
    ...
    .clk_b2(),  // was residue_clk_prbsgen
    ...
);
```

vlog/chip_src/tx_16t1/tx_top.sv
sgherbst: Thanks -- here's a short reference on SystemVerilog interfaces: https://www.doulos.com/knowhow/systemverilog/systemverilog-tutorials/systemverilog-interfaces-tutorial/.  I think that you can mostly copy & modify ``acore_debug_intf``, but this background might be helpful to better understand that code.

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: Hmmm, I didn't realize that. I will think twice about whether it is necessary to have this divider or a simpler div_b2 dff will be enough. Thanks for the note!

> I don't think gating circuitry is needed `ext_clk` and `mdll_clk`, since they go directly to the `input_divider`, which already handles clock selection and can be disabled.

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: > Please mention to @sjkim85 that you want to use his `input_divider` block. I think that he is already making some updates so that `phase_interpolator` can be used as a standalone block, so please check if similar updates are needed to use `input_divider` in the transmitter.

Hmmm, I didn't realize that. I will think twice about whether it is necessary to have this divider or a simpler div_b2 dff will be enough. Thanks for the note!

sgherbst: Yeah, I think the question is whether ``input_divider.sv`` can be run through synthesis and PnR like a "normal" digital block, or whether a bunch of special settings are required, in which case it would be better for you to get the ``input_divider`` from @sjkim85 as black box layout.  I'm not sure which is the case, so it would be best to check with him directly.

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: Thanks for sharing the helpful materials.

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: Thanks for the tutorial, this helps me clear a lot of confusion...


It appears to me that this could generate an extra DFF on the data path, which I would be a little hesitant to do so since this could be problematic in timing. 
The sequential logic of the Tx is mainly connected by explicit `wire` definitions, the DFFs in this Tx are explicitly defined and connected with other blocks...
I will go through the code with emphasis on the `cke and rst` once more.

> Two notes here:
> 
> 1. It is not allowed to `assign` to a `reg` -- you can only assign to a `wire`.  I think you could solve this problem by changing `clk_prbsgen_reg` to a `wire` (and probably change the name, too, since it is not a `reg`).  This might help to clear things up: https://www.verilogpro.com/verilog-reg-verilog-wire-systemverilog-logic/.
> 2. In general, I would advise against adding logic in the clock path like this, since it may be problematic for closing timing in synthesis, and also because it can introduce clock glitches.  I prefer to use `cke` in the sequential logic that uses the clock, e.g.
> 
> ```verilog
> always @(posedge clk) begin
>     if (rst) begin
>         q <= 0;  // reset output
>     end else if (cke) begin
>         q <= d;  // clock enabled
>     end else begin
>         q <= q;  // clock disabled; do nothing
>     end
> end
> ```
> 
> Automated clock gating in synthesis should recognize this pattern and insert clock gating circuit if it is advantageous from a power perspective. I don't have a great reference for this, but you could look at Fig. 2 and Fig. 3 in this paper to get a better sense of what I'm talking about: https://ieeexplore-ieee-org.stanford.idm.oclc.org/stamp/stamp.jsp?tp=&arnumber=5733874

sgherbst: I think the two coding styles should be logically equivalent (Fig. 2 vs. Fig. 3 in that paper), so I don't think the style I'm suggesting should cause an extra DFF to be added.  The difference is that, rather than applying clock gating logic to ``clk_prbsgen`` within ``tx_top``, you would directly wire ``clk_prbsgen`` to an output of ``tx_top``.  Then in the ``digital_core`` logic that uses ``clk_prbsgen``, I would implement sequential logic in a way that is compatible with automated clock gating.

Combining several comments, I'm basically suggesting that you remove lines 94-98 in ``tx_top`` (the three clock gating statements and the two statements related to ``residue_clk_halfrate`` and ``residue_clk_prbsgen``)

CansWang: > I think the two coding styles should be logically equivalent (Fig. 2 vs. Fig. 3 in that paper), so I don't think the style I'm suggesting should cause an extra DFF to be added. The difference is that, rather than applying clock gating logic to `clk_prbsgen` within `tx_top`, you would directly wire `clk_prbsgen` to an output of `tx_top`. Then in the `digital_core` logic that uses `clk_prbsgen`, I would implement sequential logic in a way that is compatible with automated clock gating.
> 
> Combining several comments, I'm basically suggesting that you remove lines 94-98 in `tx_top` (the three clock gating statements and the two statements related to `residue_clk_halfrate` and `residue_clk_prbsgen`)

Yeah, I will remove them. There is a little confusion from my previous comments, 
For the data path I have in the Tx top, the DFF is instantiated directly using `ff_c.sv` in `analog_core/gates`, which doesn't come with a gating/rst option. To adopt this style, I need to revise the `ff_c.sv`

Originally, I did not want to revise the ff_c.sv, so I just gate the input clock separately.

I will remove these gating codes and revise some of the `always ()` statements to follow the recommendations.

sgherbst: Sorry if I'm creating the confusion here... for the code you have right now, the gated clock signal ``clk_prbsgen_reg`` goes straight to the output of ``tx_top`` and does not appear to be used within ``tx_top``.  So I'm just saying that if you make ``clk_prbsgen`` the output of ``tx_top`` and remove the gated version, I don't think that you have to change anything else.  By the way, if you did want to create a new DFF with a clock enable, please do not modify ``ff_c``, as it is used within ``analog_core``, but instead make a copy and give the new module a different name.  But from my previous comments, I don't think this is necessary; I think you can just drive ``clk_prbsgen`` as an output of ``tx_top``, because the gating for that clock will happen in ``digital_core``.

CansWang: Noted, I will remove them.

I was trying to make sure that nothing comes in and nothing comes out by enabling the gating signal.
Sorry for the misunderstanding
Thank you for the elaborations

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,227 @@
...
+    input wire mdll_clk, // Clock from MDLL
+    input wire ext_clk, // Clock from external source
+    
+    input wire rst, // Global reset for Tx
+    input wire cke, // Gloabl lock gating for Tx
+    input wire logic [Npi-1:0] ctl_pi[Nout-1:0],
+    input wire logic clk_async,
+    output wire logic clk_encoder,
+    input wire logic ctl_valid,
+    // Control and input of the phase interpolator
```
sgherbst, line 16: I prefer to delete obsolete code rather than commenting it out. We can always restore the old code from the git history, but it can be confusing to read/review code when a lot of it is commented out. I would typically only leave in a commented block of code if there was a remark explaining it (i.e., "uncomment this code to create an interactive plot").

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,227 @@
...
 //This Tx top specify the connect between qr_4t1_mux_top and hr_16t4_mux_top
 
-//Instantiate half-rate 16 to 4 mux top
-wire [3:0] qr_data;  // Output of 16 to 4 mux
+// Instantiate half-rate 16 to 4 mux top
+wire [3:0] qr_data_p;  // Output of 16 to 4 mux, positive
+wire [3:0] qr_data_n;  // Output of 16 to 4 mux, negative
+wire residue_clk_halfrate;
+// wire residue_clk_prbsgen;
+wire residue_buf1;
```
sgherbst, line 78: `residue_buf1` and `residue_buf2` don't appear to be used anywhere

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,227 @@
...
+wire residue_buf2;
+
 wire clk_halfrate;  // Input clock for 16 to 4 mux
+wire clk_q;  // The clock inout must follow this order, (rising edge order) Q->I->QB->IB-Q
```
sgherbst, line 82: `clk_q`, `clk_i`, `clk_qb`, and `clk_ib` don't appear to be used anywhere

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,227 @@
...
+wire [3:0] clk_interp_slice; // Output from the phase interpolator
+wire [3:0] clk_interp_sw; //
+
+
+
+wire clk_in_pi;
+
+// Global reset 
+wire rstb;
+rstb = ~ rst;
```
sgherbst, line 96: I think an `assign` is needed here, i.e. `assign rstb=~rst;`

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,227 @@
...
+                .clk_out_sw(tx.clk_interp_sw[k]),
+                .Qperi(tx.Qperi[k]),
+                .cal_out(tx.cal_out_pi[k]),
+                .del_out(inv_del_out_pi[k]),
+                .pm_out(tx.pm_out_pi[k]),
+                .max_sel_mux(tx.max_sel_mux[k])
+            );
+            // portion 4 checked | Yes
+            assign tx.pi_out_meas[k] = (tx.sel_meas_pi[k] ? clk_interp_slice[k] : clk_interp_sw[k]) & tx.en_meas_pi[k];
+            assign en_unit_pi[k] = ~tx.enb_unit_pi[k];
```
sgherbst, line 152: `en_unit_pi` doesn't appear to be connected to anything -- in `analog_core`, each `en_unit_pi[k]` is wired to the `en_unit` pin of an instance of `phase_interpolator`.  is that not needed here?

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,227 @@
...
+    // input wire tx.bypass_inbuf_div2,
+    // input wire tx.en_inbuf_meas,
+    // input wire tx.sel_del_out_pi
+    // // Divider output 
+    // output wire tx.del_out,
+    // output wire tx.del_out_pi,
+
+    // output reg tx.inbuf_out_meas,
+    output wire clk_prbsgen,  // Output clock for 16-bit prbs generator
+    output reg dout_p, // Data output
```
sgherbst, line 66: I think `dout_p` and `dout_n` should both be `wire` rather than `reg`

vlog/chip_src/tx_16t1/div_b2.sv
```diff
@@ -2,6 +2,8 @@
 
 module div_b2 (
     input wire clkin,
+    input wire rst,
```
sgherbst, line 5: I don't think you have go back through your design adding `cke`, because you can already shut down the transmitter clock via the `en` signal of the `input_divider`.

vlog/chip_src/tx_16t1/div_b2.sv
```diff
@@ -10,8 +12,14 @@ initial begin
 end
 
 always@(posedge clkin) begin
-    #0.3;
+    if (rst) begin
```
sgherbst, line 15: From the previous comments, I think that you would just want this to be
```verilog
if (rst) begin
    clkout <= 0;
end else begin
    clkout <= ~clkout;
end
```
By the way, blocking (`=`) vs. `<=` (non-blocking) assignments can be confusing.  In general, if you're implementing sequential logic like this, then you should use the `<=` non-blocking assignment, while combination logic would use `=` (blocking).  There is some more information on this here: http://courses.csail.mit.edu/6.111/f2007/handouts/L06.pdf

vlog/chip_src/tx_16t1/tx_top.sv
```diff
@@ -1,38 +1,227 @@
...
-    .clk_IB(clk_ib),
-    .din(qr_data), // Quarter-rate data from half-rate 16 to 4 mux
+qr_4t1_mux_top qr_mux_4t1_0 (
+    .clk_Q(tx.clk_interp_slice[0]),  // Quarter-rate clock input
+    .clk_QB(tx.clk_interp_slice[2]),
+    .clk_I(tx.clk_interp_slice[1]),
+    .clk_IB(tx.clk_interp_slice[3]),
+    .din(qr_data_p), // Quarter-rate data from half-rate 16 to 4 mux
+    .rst(rst),
+    .cke(cke),
```
sgherbst, line 201: same as earlier comment -- I don't think you have to go back through your design to add `cke` here, because you can already shut down the TX clock via the `input_divider`

vlog/chip_src/tx_16t1/tx_debug_intf.sv
```diff
@@ -0,0 +1,232 @@
...
+		logic [$clog2(Nunit_pi)-1:0] Qperi[Nout-1:0]; 		// to JTAG
+ 		logic [$clog2(Nunit_pi)-1:0] max_sel_mux[Nout-1:0];	// to DCORE and JTAG
+       	logic [Nout-1:0] pi_out_meas;						// to output buffer
+       	//ADCrep
+       	// logic [1:0] del_out_rep;							// ([1] --> to output buffer [0] --> NC/open)
+		
+		// Input clock signal measurement
+		logic inbuf_out_meas;								// to output buffer									
+		
+	    // logic en_TDC_phase_reverse;
```
sgherbst, line 70: same comment as in `tx_top` -- please delete obsolete code rather than commenting it out, since it makes the code harder to read & review

vlog/chip_src/tx_16t1/tx_top.sv
CansWang: Oops, I missed one connection. 
`en_unit_pi` added.

sgherbst (approved):
LGTM, thanks!

PR #130, opened 2020-10-14
Remove misc_ctrl_bits register and move PRBS debug I/O to system clock

sgherbst:
This small PR does two things:
1. It replaces the ``misc_ctrl_bits`` JTAG register with single-bit registers ``cdr_en_clamp``, ``pfd_cal_flip_feedback``, ``en_pfd_cal_ext_ave``, ``en_int_dump_start``, and ``int_dump_start``.  This should make it easier to use these features.
2. PRBS debug I/O is moved from the ``Test`` to the ``System`` clock domain.  In general, we should review the JTAG clock domain assignments to make sure there are not initialization or synchronization hazards.  However, I have only done this for the PRBS and histogram circuitry because I am familiar with it.


zamyers (approved):
LGTM

PR #132, opened 2020-10-20
Top-level TX integration

sgherbst:
## Summary
This PR instantiates the transmitter at the top level of DragonPHY and wires it up.  This includes additional logic in the digital core to set the PI codes for the TX, as well as a block that can generate four kinds of TX data patterns:
1. PRBS: intended for a loopback test
2. Pulse generator: intended for channel characterization.  The width of the pulse can be 1-16 UI.
3. Constant output: intended for DC characterization of the TX output buffer.
4. Square wave generator, adjustable from 8 GHz down to about 8 kHz.  Intended for debugging the TX output bandwidth.

## Not included in this PR
1. System-level test of the TX.  This will be added after the block-level test of the TX is complete.
2. Self-checking capability in the block-level test of the TX data generator.  Will be added in a future PR.  (For now the test passes if the simulation completes without errors.)

## Other details
1. Added JTAG registers for the TX and its data generator.
2. Did a preliminary update of the synthesis constraints for the transmitter, but didn't add any constraints within the transmitter itself for the various divided clocks.  I'll need input from @CansWang on that, but it can wait until the next PR, since the synthesis test is passing.
3. Updated mflowgen steps having to do with the phase interpolator and input divider, which are instantiated as black boxes by the transmitter.
4. Wrapped ``default_nettype none`` / ``default_nettype wire`` around TX design sources to help catch typos (this will cause the simulator / synthesis tool to error out when an implicit wire is detected).  I had to update some IOs to ``input wire logic`` or ``output wire logic`` to be compatible with this change.
4. Moved ``fppi`` to the ``vlog/tb`` folder since it is not a design source.
5. Removed ``md/sram_intf_md`` (note that it ends with ``_md``, not ``.md``).  It looks like this file is obsolete.


designs/dragonphy_top/construct.py
```diff
@@ -80,6 +80,8 @@ def construct():
             Step( this_dir + '/analog_core'       ),
             Step( this_dir + '/input_buffer'      ),
             Step( this_dir + '/output_buffer'     ),
+            Step( this_dir + '/input_divider' ),
```
sgherbst, line 83: @sjkim85 note that ``input_divider`` and ``phase_interpolator`` are being provided using the tarball method, like ``analog_core`` and ``mdll_r1``

designs/dragonphy_top/construct.py
```diff
@@ -165,6 +175,8 @@ def construct():
         'analog_core.spi',
         'mdll_r1_top_macro.cdl',
         'input_buffer.spi',
+        'input_divider.spi',  # TODO: should this be *.lvs.v instead?
```
sgherbst, line 178: @sjkim85 what is the correct extension for these netlists, which are used for LVS?

dragonphy/jtag.py
```diff
@@ -11,10 +11,15 @@ def __init__(self,  filename=None, **system_values):
         id_code = self.calc_id_code()
         print(f'JTAG ID Code: 0x{id_code:08x}')
 
+        # get a list of markdown files and sort by name
```
sgherbst, line 14: I wanted to make the ordering of registers more predictable, so I am sorted the Markdown files by name.

make.py
```diff
@@ -18,19 +18,27 @@ def create_fpga_graph():
     # Dependencies
     graph.add_input('acore_intf', ext='md', folders=['md'])
     graph.add_input('cdr_intf', ext='md', folders=['md'])
-    graph.add_input('sram_multi_intf', ext='md', folders=['md'])
-    graph.add_input('sm_ffe_intf', ext='md', folders=['md'])
     graph.add_input('dcore_intf', ext='md', folders=['md'])
+    graph.add_input('hist_intf', ext='md', folders=['md'])
```
sgherbst, line 22: Not sure if this is really necessary, but I noticed that some Markdown files were missing from the build graph dependencies, so I added them.

md/dcore_intf.md
```diff
@@ -55,3 +55,10 @@
 | en_pfd_cal_ext_ave         |         |                        |                  | Test         | out      |   0       |
 | en_int_dump_start          |         |                        |                  | Test         | out      |   0       |
 | int_dump_start             |         |                        |                  | Test         | out      |   0       |
+| tx_en_ext_max_sel_mux      |         |                        |                  | Test         | out      | 'b0       |
```
sgherbst, line 58: These new registers mimic the functionality of the PI control registers for the analog core.  The main difference is that there is no MM-CDR for the TX.

md/tx_intf.md
```diff
@@ -0,0 +1,34 @@
+| Name                          | Signed? | Packed Dim             | Unpacked Dim | Clock Domain | JTAG Dir | Reset Val          |
```
sgherbst, line 1: This new set of JTAG registers is basically the PI-specific subset of registers from the analog core debug interface.

md/txdata_intf.md
```diff
@@ -0,0 +1,12 @@
+| Name                       | Signed? | Packed Dim      | Unpacked Dim   | Clock Domain | JTAG Dir | Reset Val                                    |
```
sgherbst, line 1: This set of registers controls the logic that produces data for the TX.  It has four modes: PRBS, constant, square wave, and pulse.

setup.py
```diff
@@ -38,7 +38,7 @@
     'lxml',
     'scikit-rf',
     'BeautifulSoup4',
-    'justag==0.0.4.2',
+    'justag==0.0.4.4',
```
sgherbst, line 41: updated the JusTAG version to resolve an issue having to due with a limit on the total number of registers.

tests/cpu_block_tests/tx_data_gen/test.sv
```diff
@@ -0,0 +1,58 @@
+`timescale 1fs / 1fs
```
sgherbst, line 1: This test isn't yet self-checking.  But it is still useful as a manual check on the TX data generator -- it is easy to verify the four data generation modes by eye when viewing the waveforms.  This shouldn't be too hard to automate in a future PR.

vlog/chip_src/digital_core/digital_core.sv
```diff
@@ -250,6 +267,22 @@ module digital_core import const_pack::*; (
         end
     endgenerate
 
+    // for tx_top
```
sgherbst, line 270: The arithmetic here should match the arithmetic for the analog core PIs, except there is not MM-CDR

vlog/chip_src/digital_core/digital_core.sv
```diff
@@ -567,6 +602,34 @@ module digital_core import const_pack::*; (
...
     assign clk_cgra = (clk_adc & en_cgra_clk_latch);
 
+    // transmitter control signals
+
+    assign tx_rst = ddbg_intf_i.tx_rst;
+    assign tx_ctl_valid = ddbg_intf_i.tx_ctl_valid;
+
+    // TX data generator
+
+    tx_data_gen #(
```
sgherbst, line 612: this is the new block that can produce PRBS, square, pulse, etc. data for the TX

vlog/chip_src/top/dragonphy_top.sv
```diff
@@ -11,6 +11,11 @@ module dragonphy_top import const_pack::*; (
 	input `pwl_t ext_rx_inp_test,
 	input `pwl_t ext_rx_inn_test,
 
+    // analog outputs
+    // TODO: use pwl_t
```
sgherbst, line 15: once the TX buffer implementation is determined, we should switch the data type of the TX outputs to ``pwl_t``, as is used for the RX inputs.

vlog/chip_src/top/dragonphy_top.sv
```diff
@@ -133,8 +164,39 @@ module dragonphy_top import const_pack::*; (
...
 		.adbg_intf_i(adbg_intf_i)            // debug IO
 	);
-	
-	// digital core instantiation
+
+    ///////////////////////////////
+	// transmitter instantiation //
+	///////////////////////////////
+
+    tx_top itx (
```
sgherbst, line 172: this is the top-level instantiation of the transmitter

vlog/chip_src/tx_16t1/qr_4t1_mux_top.sv
```diff
@@ -39,16 +41,11 @@ ff_c dff_IB0 (.D(din[3]), .CP(clk_Q), .Q(D0DIB)); // data captured using Q clk a
...
-//wire [1:0] sel;
+// 4 to 1 mux
 div_b2 div (.clkin(clk_IB), .rst(rst), .clkout(ck_b2));
-// Combinational logic to generate the selection window from clk Q-QB-I-IB
-// assign sel =  
 
-always @(D1MQ or D1MI or D2MQB or D2MIB or clk_Q or clk_QB or clk_I or clk_IB) begin
-    assign data = (clk_Q && clk_I) ? D1MQ : ((clk_I && clk_QB) ? D1MI : ((clk_QB && clk_IB) ? D2MQB : D2MIB)); // Maybe problematic to write it in this way, need to check after synthesis
-end
+assign data = (clk_Q && clk_I) ? D1MQ : ((clk_I && clk_QB) ? D1MI : ((clk_QB && clk_IB) ? D2MQB : D2MIB));
```
sgherbst, line 47: I had to remove the ``always`` block because it was causing a synthesis error.  This should be equivalent to what was there before.

vlog/fpga_models/tx_16t1/tx_top.sv
```diff
@@ -0,0 +1,38 @@
+`default_nettype none
```
sgherbst, line 1: for now, the TX is basically ignored in emulation.  this can be addressed in a future PR.

vlog/chip_src/top/dragonphy_top.sv
```diff
@@ -133,8 +164,39 @@ module dragonphy_top import const_pack::*; (
...
+        .ext_clk(clk_main),
+
+        .rst(tx_rst),
+        .ctl_pi(tx_pi_ctl),
+        .clk_async(clk_async),
+
+        ////////////////////////////////////////////////
+        // TODO: review clk_encoder for TX.  should it
+        // be derived from one of the PIs in the TX?
+        .clk_encoder(clk_adc),
```
sgherbst, line 184: @sjkim85 what do you think ``clk_encoder`` for the TX PIs should be connected to?  I don't think it matters when the TX PI control codes are set completely externally, but when they are "normalized" using the ``max_sel_mux`` signals, this may be important to make sure a stable PI code is sampled.

vlog/chip_src/tx_16t1/qr_4t1_mux_top.sv
CansWang: Well noted, thanks for revising this

designs/dragonphy_top/construct.py
sjkim85: I made the tarballs and sent the path of the files to @zamyers.
Now the tarballs only contain LEF but I will update them soon.

designs/dragonphy_top/construct.py
sjkim85: @sgherbst , .spi is correct for both phase_interpolator and input_divider.

vlog/chip_src/top/dragonphy_top.sv
sjkim85: @sgherbst, I agree. I think clk_adc (or system clock) is fine.

sjkim85 (approved):

PR #138, opened 2020-10-25
prbs test & fixed pattern test added!

CansWang:
Fixed pattern and prbs test test added

### DIFF
--- a/tests/cpu_block_tests/tx_top_sim/test.sv
+++ b/tests/cpu_block_tests/tx_top_sim/test.sv
@@ -408,6 +408,7 @@ end
         end
     end
 
+    // Counter initialization
     always @(posedge clk_prbs) begin
         if (rst) begin
             count_flag[0] <= 8'b0;
@@ -419,7 +420,7 @@ end
             count_flag[6] <= 8'b0;
             count_flag[7] <= 8'b0;
         end else if (record_flag[0]) begin
-            #0.875ns;
+            #0.875ns;  // Manually align the count down with the first bit of the data output, this delay should be fixed if not circuit connection is changed
             count_flag[0] = 8'd16; // determine how many bits to store in the shift reg
             record_flag[0] = 1'b0;
         end else if (record_flag[1]) begin
@@ -457,6 +458,13 @@ end
     always @(posedge clk_oversample) begin
         if (rst) begin
             parecord1 <= 16'h0000;
+            parecord2 <= 16'h0000;
+            parecord3 <= 16'h0000;
+            parecord4 <= 16'h0000;
+            parecord5 <= 16'h0000;
+            parecord6 <= 16'h0000;
+            parecord7 <= 16'h0000;
+            parecord8 <= 16'h0000;
         end else if ((!prbs_den) && (count_flag[0] > 0)) begin
             //shift register
             parecord1 <= {dout_p, parecord1[15:1]};  // Right shift to store the 16-bit data package in the correct order
--- a/tests/cpu_block_tests/tx_top_sim/test.sv
+++ b/tests/cpu_block_tests/tx_top_sim/test.sv
@@ -82,6 +82,7 @@ tx_top tx_mux (
     .clk_async(clk_async),
     .clk_encoder(clk_encoder),
     .ctl_valid(ctl_valid),
+    .rst(rst),
     .clk_prbsgen(clk_prbs),  // Output clock for 16-bit prbs generator
     .dout_p(dout_p),
     .dout_n(dout_n),
@@ -96,28 +97,22 @@ div_b2 divb2 (.clkin(clk_oversample), .rst(rst), .clkout(clk_prbschecker));
 // fppi pi(.clkin(clk_a), .clk_Q(cq), .clk_I(ci), .clk_QB(cqb), .clk_IB(cib));
 // div_b2 div0(.clkin(clk_full), .clkout(clk_a));
 
-
 initial begin
 
     `ifdef DUMP_WAVEFORMS
 	        $shm_open("waves.shm");
 	        $shm_probe("ASMC");
     `endif
-
-    // clk_a  = 1'b0;
     // Initialize all the nodes
-    //Global
+    // Global
     clk_full = 1'b0;
     clk_2 = 1'b0;
     clk_oversample = 1'b0;
     // clk_encoder =1'b0; 
     rst = 1'b1;
-    tx_intf.rst = 1'b1;
-    tx_intf.cke = 1'b1;
     cke = 1'b1;
     rst_prbs = 1'b1;
     #3ns;
-
     // Input divider
     tx_intf.en_inbuf = 1'b0;
     tx_intf.sel_clk_source = 1'b0;
@@ -192,8 +187,6 @@ initial begin
     #5ns; // After 50 units time, release the enable and reset button
     rst = 1'b0;
     cke = 1'b1;
-    tx_intf.rst = 1'b0;
-    tx_intf.cke = 1'b1;
     // PI setting
     #5ns;
     tx_intf.en_inbuf = 1'b1;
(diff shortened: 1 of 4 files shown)


vlog/chip_src/tx_16t1/qr_4t1_mux_top.sv
```diff
@@ -44,7 +44,7 @@ ff_c dff_IB2 (.D(D1DIB), .CP(clk_QB), .Q(D2MIB));
 // 4 to 1 mux
 div_b2 div (.clkin(clk_IB), .rst(rst), .clkout(ck_b2));
 
-assign data = (clk_Q && clk_I) ? D1MQ : ((clk_I && clk_QB) ? D1MI : ((clk_QB && clk_IB) ? D2MQB : D2MIB));
+assign data = (clk_Q && clk_I) ? D1MQ : ((clk_I && clk_QB) ? D1MI : ((clk_QB && clk_IB) ? D2MQB : ((clk_IB && clk_Q) ? D2MIB : data)));
```
CansWang, line 47: If the clock window does not belong to the four cases, maintain the current value.

tests/cpu_block_tests/tx_top_sim/test.sv
```diff
@@ -177,34 +213,294 @@ initial begin
...
-    rst_prbs = 1'b0;
-    #8ns;
+    $display("Release prbs generator reset..."); // The reset of the prbs must hold unit valid clock are generated and fed into TX
+    rst_prbs = 1'b0;  // Otherwise the initall value will not be loaded
+    prbs_den = 1'b1;
+    #5ns;  // wait 5 ns to avoid initial state error
+    
+    
+    
+    // Test case 1 | PRBS test
```
sgherbst, line 244: Rather than running the PRBS test and checking for a specific number of error bits, I suggest running the PRBS test without error injection, and then check for zero error bits.  (For your own testing, you can inject errors and check that the test fails, but that doesn't have to be in the regression suite, because we already have such tests for the PRBS generator and checker.)

CansWang: > Rather than running the PRBS test and checking for a specific number of error bits, I suggest running the PRBS test without error injection, and then check for zero error bits. (For your own testing, you can inject errors and check that the test fails, but that doesn't have to be in the regression suite, because we already have such tests for the PRBS generator and checker.)

Sure, just modified it.

tests/cpu_block_tests/tx_top_sim/test.sv
```diff
@@ -177,34 +213,294 @@ initial begin
...
+    inj_error = 1'b0; 
+    #30ns;  // for the counter to finish counting 
+    error_bits_1 = err_count;
+    //store the error count
+    prbs_den = 1'b0;
+    #30ns;
+
+    
+    
+    // Test Case 2 fixed pattern
```
sgherbst, line 261: This is a recurring pattern in your test, so a recommend bundling it into a ``task`` that you can call multiple times.  For example:
```verilog
task run_test(input [15:0] data_in, input integer case_no);
    fixed_pa = data_in;
    prbs_den = 1'b0;
    //wait for 2 ns
    #1.1ns; // record flag on
    record_flag = 1'b1;
    #3ns; // stop recording
    $display("parecord: %b", parecord);
    if (parecord == data_in) begin
        $display("Pattern detected, Success! (case %0d)", case_no);
    end else begin
        $error("Pattern lost (case %0d)", case_no);
    end
endtask
initial begin
   ...
   run_test(16'b1010101010101010, 2);
   run_test(16'b1010101001010101, 3);
   ...
end
```

tests/cpu_block_tests/tx_top_sim/test.sv
```diff
@@ -177,34 +213,294 @@ initial begin
...
+            inj_error <= 0;
+            error_bits_1 <= 32'b0;
+            err_count <= 32'b0;
+        end else if (err_flag && prbs_den) begin
+            err_count += 1'b1;
+        end
+    end
+
+    // Counter initialization
+    always @(posedge clk_prbs) begin
```
sgherbst, line 413: I think you could simplify the code to read the TX bits by using a ``task``.  For example:
```verilog
task read_tx_data(output [15:0] parecord);
    integer i;
    parecord = 0;
    for (i=0; i<16; i=i+1) begin
        @(posedge clk_oversample);  // this means "wait until a positive edge of clk_oversample"
        parecord <= {dout_p, parecord[15:1]};
    end
endtask
```
You can then call this task like this:
```
logic [15:0] tx_out_1;
logic [15:0] tx_out_2;
...
read_tx_data(tx_out_1);  // 16 bits will be stored in tx_out_1
read_tx_data(tx_out_2);  // the next 16 bits will be stored in tx_out_2
...
```
You can call a ``task`` within another ``task``, so you could use ``read_tx_data`` in the ``run_test`` task I described earlier.

tests/cpu_block_tests/tx_top_sim/test_tx_top_sim.py
```diff
@@ -11,5 +11,5 @@ def test_sim():
     DragonTester(
         ext_srcs=deps,
         directory=BUILD_DIR,
-        # dump_waveforms=True,
+        dump_waveforms=True,
```
sgherbst, line 14: If you write ``dump_waveforms=dump_waveforms``, and then change the ``def test_sim():`` line to ``def test_sim(dump_waveforms)`` (i.e., adding an argument called ``dump_waveforms``), then you can enable waveform dumping from the command line when you run ``pytest``, e.g.
```
pytest -s --dump_waveforms
```
This works because of the way ``pytest`` is configured in ``conftest.py`` (where the custom argument ``dump_waveforms`` is defined)

tests/cpu_block_tests/tx_top_sim/test.sv
```diff
@@ -22,6 +22,32 @@ wire clk_encoder;
...
 logic clk_prbschecker;
+logic inj_error;
+logic prbs_den;
+logic [15:0] fixed_pa;
+logic [15:0] data;
+logic [7:0] count_flag[7:0];
+
+assign data = prbs_den ? prbsdata : fixed_pa;
+logic [7:0] record_flag;
+logic [15:0] parecord1;
```
sgherbst, line 33: I think you probably don't need to store each 16-bit word in a separate variable; see comments below on the ``run_test`` ``task``.

tests/cpu_block_tests/tx_top_sim/test.sv
```diff
@@ -177,34 +213,294 @@ initial begin
...
     ctl_pi[1] = 9'd67;
     ctl_pi[2] = 9'd135;
     ctl_pi[3] = 9'd202;
-
+    //
     ctl_valid = 1'b1;
     #5ns; // After 50 units time, release the enable and reset button
+    $display("Releasing reset...");
     rst = 1'b0;
     cke = 1'b1;
```
sgherbst, line 226: It doesn't look like ``cke`` is connected to anything.  I think this variable could be removed.

vlog/chip_src/tx_16t1/qr_4t1_mux_top.sv
sgherbst: Unfortunately this creates a combination logic loop.  We generally try to avoid this because it creates issues in timing analysis (for example, as mentioned by Intel [here](https://www.intel.com/content/www/us/en/programmable/quartushelp/13.0/mergedProjects/verify/da/comp_file_rules_loop.htm) in the context of FPGA synthesis).  What happened when the code was written as before, without this extra statement?  In general, I think it would be helpful if you could draw the hardware that you want to be generated by this line, so that we can figure out how to implement it in Verilog.  I think this muxing operation is probably one of the most critical in the TX design.

tests/cpu_block_tests/tx_top_sim/test.sv
sgherbst: Here are some reference material on Verilog ``tasks``: http://www.asic-world.com/verilog/task_func1.html

sgherbst (approved):
LGTM.  I'll add an issue to track the combo logic loop, which can be addressed in a future PR.

PR #142, opened 2020-10-28
Physical Design Flow + Datapath Updates

zamyers:
This branch is a continuation of the new_datapath branch with updates related to the physical design flow.

There are two main changes from Master in the system design:
1. The previous DSP_backend has been replaced with a new digital backend that consists of a FFE, a slicer, an aligner, a channel filter and a sliding two bit error detector. The two bit sliding error detector differs significantly from the previous error checker.
2. There is now an "error tracker" block that uses prbs_flags to sample three frames of data into a SRAM.

There are many minor changes, but the main change that touch blocks outside of the datapath is a change to the PRBS checker. The PRBS checker now outputs its error flags.

There are additional changes to the JTAG register space (as the error tracker requires its own registers).

On the Physical Design Flow, 
The two primary changes are (1) a new floorplan that includes new macros required by the TX and error tracker blocks and some minor tweaks related to register retiming.

I will also add the macro required by the histogram blocks but that will occur on the phys_design_v2 branch.


designs/dragonphy_top/synopsys-dc-synthesis/configure.yml
```diff
@@ -70,7 +71,7 @@ postconditions:
 
   # Basic error checking
 
-  - assert 'error' not in File( 'logs/dc.log' )
+  - assert 'Error:' not in File( 'logs/dc.log' )
```
sgherbst, line 74: Just wondered what issue motivated this change.  I see that the latest version of **mflowgen** is using the error string ``error:`` (lowercase ``e``), so I wanted to make sure we're still catching all the errors we used to catch.
ref: https://github.com/cornell-brg/mflowgen/blob/f646da41b57b7894fc4e8ef852eb0cd713e97cb3/steps/synopsys-dc-synthesis/configure.yml#L93

zamyers:
I changed it because I have a block that uses the word error in it.

When looking at actual errors, I found that the DC reports errors with the tag “Error:”  (the assert ‘error’ appears to be case insensitive)
So this was the simplest solution to avoid having to rewrite a lot of verilog code.

Zach

On Oct 28, 2020, at 4:34 PM, Steven Herbst <notifications@github.com<mailto:notifications@github.com>> wrote:



@sgherbst commented on this pull request.

In designs/dragonphy_top/synopsys-dc-synthesis/configure.yml<https://github.com/StanfordVLSI/dragonphy2/pull/142#discussion_r513822973>:

> @@ -70,7 +71,7 @@ postconditions:

   # Basic error checking

-  - assert 'error' not in File( 'logs/dc.log' )
+  - assert 'Error:' not in File( 'logs/dc.log' )


Just wondered what issue motivated this change. I see that the latest version of mflowgen is using the error string error: (lowercase e), so I wanted to make sure we're still catching all the errors we used to catch.
ref: https://github.com/cornell-brg/mflowgen/blob/f646da41b57b7894fc4e8ef852eb0cd713e97cb3/steps/synopsys-dc-synthesis/configure.yml#L93

sgherbst:
@zamyers could you move the three files in ``vlog/chip_src_v2`` to a new subfolder in ``tests/cpu_block_tests``?  They're not being included in the regression suite in their current location.  (I'm suggesting the ``cpu_block_tests`` folder because this test does not instantiate ``dragonphy_top``; it's lower-level).  At that point, the ``vlog/chip_src_v2`` folder should be empty, and you can delete it.  Thanks!

sgherbst:
Took a quick look at the tests -- seems that ``mm_cdr`` and ``mm_cdr_slew`` are failing (CPU tests) and the emulation ``test_2`` is failing (it's similar to ``mm_cdr``).  The rest of the emulation tests (``test_3``-``test_6`` didn't run due to the ``-x`` argument in ``pytest``).

vlog/chip_src/datapath_core/datapath_core.sv
```diff
@@ -0,0 +1,352 @@
...
+                for(jj=0; jj<ffe_gpack::length; jj=jj+1) begin
+                    weights[jj][ii] <= dsp_dbg_intf_i.weights[ii][jj];
+                    disable_product[jj][ii] <= dsp_dbg_intf_i.disable_product[jj][ii]; //Packed to Unpacked Conversion I think requires this
+                end
+            end
+    end
+
+    generate
+        for(gi = 0; gi < constant_gpack::channel_width; gi = gi + 1) begin
+            assign estimated_bits_out[gi]   = estimated_bits[gi][ffe_pipeline_depth];
```
sgherbst, line 94: It looks like this assigns single-bit values to each entry in ``estimated_bits_out[gi]``, even though the elements in that array have a bit width ``output_precision``.  Is that expected?  This may be related to one of the problems I'm seeing on this branch, where ``din`` of the MM CDR is zero.

sgherbst:
@zamyers I just pushed two small design changes to this branch:
1. Update assignment of ``estimated_bits_out`` (I think this was a typo, but am not familiar with the internal design of the DSP core)
2. Add a reset condition for ``phase_error_q`` in the MM CDR.
The tests that were failing on the regression servers now pass locally for me.  Could you confirm if these two changes are OK?  Thanks.

sgherbst (approved):
LGTM, will approve & merge.

PR #144, opened 2020-10-29
Update clock constraints for TX

sgherbst:
## Summary

This PR updates the timing constraints for signals related to the TX.  It also removes a non-synthesizable ``initial`` block from ``div_b2``.

## Details

1. ``create_clock`` is used to declare four 4 GHz clocks at the outputs of the PIs in the TX.  They are given phase shifts of 0, 90, 180, and 270 degrees.
2. There are four clock dividers in the TX, and ``create_generated_clock`` is used to describe their behaviors.
3. The setup/hold clock uncertainty for all of these clocks (4 GHz, 2 GHz, and 1 GHz) are set to a few percent their period: The 1 GHz clocks are given an uncertainty of 30 ps, the 2 GHz clocks are given an uncertainty of 20 ps, and the 4 GHz clocks are given an uncertainty of 10 ps.
4. The TX clocks are grouped together with ``set_clock_groups``.  This means that clock domain crossings between the various TX clocks are analyzed, but crossings to other clocks (``clk_adc``, ``clk_jtag``, etc.) are not.
5. The maximum transition of TX clocks is set to 10% of their periods: 100ps for 1 GHz clocks, 50ps for 2 GHz clocks, and 25ps for 4 GHz clocks.
6. ``create_clock`` is not longer used for debug clock signals that are monitored with the ``output_buffer``.  Those nets are now treated as regular signals, but given a stricter-than-default ``max_transition`` of 25ps (i.e., 10% of the 4 GHz period)

### DIFF
--- a/designs/dragonphy_top/constraints/gen_constraints.py
+++ b/designs/dragonphy_top/constraints/gen_constraints.py
@@ -7,6 +7,7 @@ design_name = os.environ['design_name']
 time_scale = float(os.environ['constr_time_scale'])
 cap_scale = float(os.environ['constr_cap_scale'])
 main_per = float(os.environ['constr_main_per'])
+clk_4x_per = 0.25*main_per*time_scale
 
 output = ''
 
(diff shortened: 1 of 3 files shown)


vlog/chip_src/tx_16t1/div_b2.sv
```diff
@@ -8,11 +6,7 @@ module div_b2 (
     output reg clkout
 );
 
-initial begin
```
sgherbst, line 11: ``initial`` is not synthesizable.  However, if I simply remove it, the ``tx_loopback`` test fails.  The reason is that clock starts toggling exactly when ``rst`` goes low, so the reset state is never encountered.  For now, I fixed this by changing ``rst`` to an asynchronous reset.  However, we need to review this to make sure that the reset scheme is actually robust -- it's possible for the clock dividers on the positive and negative side of the TX data flow to start up with opposite polarities.

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -32,19 +33,83 @@
...
-# TODO: add generated clocks in TX
+#############
+# TX clocks #
+#############
 
-create_clock -name clk_prbsgen \\
-    -period {main_per*time_scale} \\
-    -waveform {{0 {0.5*main_per*time_scale}}} \\
-    [get_pins itx/clk_prbsgen]
+# PI outputs
```
sgherbst, line 40: This is where the four PI outputs are declared as clock signals (with 0, 90, 180, and 270 degrees phase shift)

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -32,19 +33,83 @@
...
+    -waveform {{{0.5*clk_4x_per} {clk_4x_per}}} \\
+    [get_pins itx/iPI[2].iPI/clk_out_slice]
+
+create_clock -name clk_tx_pi_3 \\
+    -period {clk_4x_per} \\
+    -waveform {{{0.75*clk_4x_per} {1.25*clk_4x_per}}} \\
+    [get_pins itx/iPI[3].iPI/clk_out_slice]
 
-# clock uncertainty
+# clock dividers
```
sgherbst, line 62: This is where the behavior of the internal clock dividers in the TX is declared.

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -32,19 +33,83 @@
...
+#####################
+# clock uncertainty #
+#####################
+
+# clk_retimer
 set_clock_uncertainty -setup 0.03 clk_retimer
 set_clock_uncertainty -hold 0.03 clk_retimer
-set_clock_uncertainty -setup 0.03 clk_prbsgen
-set_clock_uncertainty -hold 0.03 clk_prbsgen
+
```
sgherbst, line 91: Clock uncertainties set to a few percent of the clock period.

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -86,7 +151,16 @@
 
 set_clock_groups -asynchronous \\
     -group {{ clk_jtag }} \\
-    -group {{ clk_prbsgen }} \\
+    -group {{ \\
```
sgherbst, line 154: Inter-clock timing of TX clocks is analyzed, but their timing with respect to JTAG and ADC clocks is not.

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -125,6 +199,8 @@
     ext_rstb \\
     ext_dump_start \\
     clk_cgra \\
+    ramp_clock \\
```
sgherbst, line 202: These are two new pins for the CDR (i.e., unrelated to the TX)

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -216,21 +292,24 @@
 # Tighten transition constraint for clocks declared so far
 set_max_transition {0.1*time_scale} -clock_path [get_clock clk_jtag]
 set_max_transition {0.1*time_scale} -clock_path [get_clock clk_retimer]
-set_max_transition {0.1*time_scale} -clock_path [get_clock clk_prbsgen]
+set_max_transition {0.025*time_scale} -clock_path [get_clock clk_tx_pi_0]
```
sgherbst, line 295: The transition time for TX clocks is set to 10% of their period

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -216,21 +292,24 @@
...
+set_max_transition {0.1*time_scale} -clock_path [get_clock clk_tx_qr_1]
 
 # Set transition time for high-speed signals monitored from iacore and itx
-# transition time is 10% of a 4 GHz period.  Note that we have to
-# create a clock before the transition constraint is applied
-# (appears to be related to the fact that these are internal nets)
+# transition time is 10% of a 4 GHz period.
 
 foreach x [get_object_name $adbg_clk_pins] {{
-    create_clock -name clk_mon_net_$x -period {0.25*time_scale} [get_pins $x]
```
sgherbst, line 227: No longer using ``create_clock`` for clock signals that are only for debugging purposes.

vlog/chip_src/tx_16t1/div_b2.sv
CansWang: Thanks for the note, 

Yes, this could be problematic

designs/dragonphy_top/constraints/gen_constraints.py
CansWang: Does this `uncertainty` specify the clock jitter?
What is the value stands for, rms/pk-pk?

thanks for adding this.

> Clock uncertainties set to a few percent of the clock period.

CansWang (approved):

PR #146, opened 2020-10-30
output_buf module create

CansWang:


CansWang:
Just create the output buffer module.

1) Dividers `div_b2` that were originally in `qr_4t1` and `hr_16t4` are moved to `tx_top`.
2) Interface added for output buffer slice number control. Control bits are 8-bit wide and independent between positive and negative output.

At this point, the buildkite check will definitely fail because the output buffer is a direct instantiation of tsmcN16 std cell logic gate.
Please ignore this temporary. @sgherbst

vlog/chip_src/tx_16t1/tx_debug_intf.sv
```diff
@@ -47,8 +47,9 @@ interface tx_debug_intf import const_pack::*; (
 		// Input clock signal measurement
 		logic inbuf_out_meas;								// to output buffer									
 		
-		
-		
+		// output buffer control
+        logic [7:0] ctl_buf_n;
```
sgherbst, line 51: @CansWang thanks for adding these debug signals.  Here's how to integrate them into the JTAG interface:
1. Open ``dragonphy2/md/tx_intf.md`` and create entries for both signals.  The packed dimension for both is ``7:0`` and the unpacked dimension can be left blank.  (This is where you get to specify the default values)
2. Open ``dragonphy2/vlog/chip_src/jtag/jtag.sv`` and wire from ``rjtag_intf_i`` (which contains all JTAG signals) to ``tdbg_intf_i`` (which contains only TX signals).  This is where the TX-related wiring is done:
https://github.com/StanfordVLSI/dragonphy2/blob/e10cc1037cf821ebd21629ee69186df9a63261d5/vlog/chip_src/jtag/jtag.sv#L317-L346

vlog/chip_src/tx_16t1/output_buf_tx.sv
```diff
@@ -0,0 +1,60 @@
+module output_buf_tx #(
```
sgherbst, line 1: Could you add a behavioral model for the output buffer?  That should get tests passing again.  You can do this by creating a file ``vlog/cpu_models/tx_16t1/output_buf_tx.sv`` and just have it wire from ``DINN`` to ``DOUTN`` and from ``DINP`` to ``DOUTP``.  (We can add more detail later).  The dependency system should automatically figure out to use that model for CPU-based simulations.

Once that simulation is working, please copy the file to ``vlog/fpga_models/tx_16t1/output_buf_tx.sv`` so that it can be used in FPGA emulation.  (I may update it later, but this should at least get tests passing again)

vlog/chip_src/tx_16t1/output_buf_tx.sv
```diff
@@ -0,0 +1,60 @@
...
+	output wire logic DOUTN,
+    output wire logic DOUTP
+);
+	// Internal connections
+    wire logic BTN; // Buffer to Termination -
+    wire logic BTP; // Buffer to Termination +
+
+
+    // Termination 
+    `ifndef VIVADO
```
sgherbst, line 17: You don't need the ``ifndef``/``endif`` in this case because we're replacing the output buffer with a model for FPGA emulation

vlog/chip_src/tx_16t1/output_buf_tx.sv
```diff
@@ -0,0 +1,60 @@
...
+            .VinN(DOUTP),
+            .Vcm(BTP)
+        );
+
+    `endif
+    
+    // instantiate BUFTD +
+    genvar i;
+    generate 
+        for (i=0; i<8; i=i+1) begin: iBUF
```
sgherbst, line 36: please call label this generate block with something like ``iBUFP`` and the other one with something like ``iBUFN``; otherwise it will be hard to tell which buffers are ``P`` and which are ``N`` (might need to know this for PnR scripts)

vlog/chip_src/tx_16t1/output_buf_tx.sv
```diff
@@ -0,0 +1,60 @@
...
+	        BUFTD4BWP16P90 tri_buf (
```
sgherbst, line 37: For the tristate buffer, can you try this procedure?
1. Instantiate it as something like ``tx_tri_buf`` (i.e., generic gate name)
2. Create an implementation of ``tx_tri_buf`` in ``dragonphy2/vlog/chip_src_tsmc16/tx_16t1/tx_tri_buf ``.  That implementation should instantiate the gate ``BUFTD4BWP16P90``.  In other words, it is a wrapper for the process-specific gate.
3. Then, create an implementation of ``tx_tri_buf`` in ``dragonphy2/vlog/chip_src_freepdk45/tx_16t1/tx_tri_buf ``.  That implementation should instantiate an equivalent cell from FreePDK45 -- the options are ``TBUF_Xn``, where ``n`` is ``1``, ``2``, ``4``, ``8``, or ``16``.  Note that for the FreePDK45 cells, the input is called ``A``, the output enable is called ``EN``, and the output is called ``Z``.  The reason that we need the FreePDK45 version is because that is what our regression server uses for synthesis tests.
3. Finally, update our dependency management system so that it uses the right view for TSMC16 vs. FreePDK45.  Open ``dragonphy2/dragonphy/views.py`` and edit this section:
https://github.com/StanfordVLSI/dragonphy2/blob/e10cc1037cf821ebd21629ee69186df9a63261d5/dragonphy/views.py#L155-L166

In the ``freepdk-45nm`` section, add
```python
override['tx_tri_buf'] = 'chip_src_freepdk45'
skip.add('TBUF_Xn')  # use your value of n (1, 2, 4, 8, or 16)
```

Similarly, in the ``tsmc16`` section, add
```python
override['tx_tri_buf'] = 'chip_src_tsmc16'
skip.add('BUFTD4BWP16P90')
```

vlog/chip_src/tx_16t1/hr_16t4_mux_top.sv
```diff
@@ -25,7 +24,7 @@ generate  // Instantiate 4 hr_4t1_mux_top to form 16:4 mux
 endgenerate
 
 // Clock divider, divide-by-two
```
sgherbst, line 26: very minor comment -- please delete obsolete code rather than commenting it out

sgherbst:
@CansWang one last comment -- it looks like this branch is out-of-date with the master branch; would you mind updating it?  Thanks!

sgherbst (approved):
LGTM

PR #147, opened 2020-11-02
36 BUFTD4 added as the output stage, dependency problem solved.

CansWang:


sgherbst (approved):
LGTM.  As I mentioned earlier, in a future PR you could concatenate two 18-bit control signals to make a single 36-bit control signal, allowing you to reduce from 4 ``generate`` blocks to 2; you could also use a single ``genvar`` declared outside of the ``generate`` statement, rather than ``i``, ``ii``, ``j``, ``jj``.  But those are small changes, and they do not affect the functionality of the block, so I will go ahead and merge.

Issue #131, opened 2020-10-19
JTAG register map is not deterministic


sgherbst:
Running `make.py` multiple times with the same JTAG inputs (i.e., Markdown files and the constant package) produces different register maps, which can be verified by diffing the file `build/all/jtag/reg_list.json` from different runs of the script.  I think we should get to the bottom of this before tapeout because it could make it quite difficult to know for sure what the register map is on the final chip.  In particular, even though we could read the Git hash as part of the JTAG ID, that information would not be sufficient to verify the recreate the register map.

sgherbst:
My guess is that this is related to the non-deterministic order of a `set` in Python3.  For example, if you run this script multiple times, you will get different results:
```python
# ref: https://stackoverflow.com/questions/60942487/non-deterministic-order-of-python-set?noredirect=1&lq=1
colors = ['blue', 'red', 'green', 'yellow']
print(set(colors))
```

I notice that a `set` is used in at least two places in JustTAG, so it could be that the non-deterministic ordering of Markdown files is changing the register address assignment.
https://github.com/StanfordVLSI/JusTAG/blob/6153254f2d7fe466edd411bfb915fd8cfd6c57e5/justag/JusTAG.py#L45-L46

Since Python 3.7, dictionaries (but not sets) preserve insertion order.  You can use this fact to create an ordered-set-like object from a dictionary:
```python
colors = ['blue', 'red', 'green', 'yellow']
print({c: None for c in colors})
```

sgherbst:
An even simpler option might be to have JusTAG depend on the ``ordered_set`` package, and switch the ``set`` instances to ``OrderedSet``.  (https://pypi.org/project/ordered-set/)

PR #149, opened 2020-11-04
Fix non-deterministic register map

sgherbst:
## Summary
This small PR resolves issue #131 by making the JTAG register map deterministic.  

## Details
1. Updates **justag** to **0.0.4.5**, which fixes the underlying issue.
2. Enforces the use of **justag** **0.0.4.5** (i.e., will not generate a register map if the version is not right)
3. Adds some code in the **experiments** directory that can be used to check if the register map generation process is repeatable.


zamyers (approved):
Fairly straight forward. 

LGTM

PR #150, opened 2020-11-06
Fix TX synthesis bug

sgherbst:
## Summary
I examined the synthesis output (for FreePDK45) and noticed a big problem -- most of the TX mux and output driver was missing.  This PR aims to fix that by applying ``dont_touch`` to some nets and cells in the transmitter.

There was also an unrelated problem, where the synthesis tool did not connect ``termination`` resistors to the DragonPHY output pins, because the ``*.lib`` file for the resistor had all pins marked as ``input``.  For now, I marked ``VinP`` and ``VinN`` as ``input`` and ``Vcm`` as ``output`` (marking all as ``inout`` caused a different issue having to do with combo loop detection).

Of course, someone with process access will need to verify these changes on the real process to be used for tapeout, to make sure that these changes fix the problem.

## Details
1. Per feedback from @zamyers: The transition time constraint for 16 GHz outputs of the TX mux is relaxed to 8ps from 6.25ps.  In addition, the target clock period is increased from 0.7 ns to 0.9 ns.
2. I noticed that when ``set_max_transition`` was being applied to pins of black boxes (analog core, MDLL, etc.), this caused a warning, and the constraint was not applied.  This PR does not solve that problem, but it at least comments out the problematic commands.
3. It looks like there is still a problem with the tool inserting a buffer after the termination resistor.  That will have to be solved in another PR.

### DIFF
--- a/designs/dragonphy_top/constraints/configure.yml
+++ b/designs/dragonphy_top/constraints/configure.yml
@@ -18,7 +18,7 @@ parameters:
 
   # Period of the main clock in nanoseconds
   # (will be scaled by constr_time_scale)
-  constr_main_per: 0.7
+  constr_main_per: 0.9
 
   # Scale factor for timing constraints
   constr_time_scale: 1.0
--- a/designs/dragonphy_top/constraints/gen_constraints.py
+++ b/designs/dragonphy_top/constraints/gen_constraints.py
@@ -274,6 +274,8 @@ for {{set i 0}} {{$i < 2}} {{incr i}} {{
 for {{set i 0}} {{$i < 2}} {{incr i}} {{
     set_dont_touch [get_cells "itx/buf1/iBUF[$i].i_tri_buf_n/tri_buf"]
     set_dont_touch [get_cells "itx/buf1/iBUF[$i].i_tri_buf_p/tri_buf"]
+    set_false_path -through [get_pins -of_objects "itx/buf1/iBUF[$i].i_tri_buf_n/tri_buf"]
+    set_false_path -through [get_pins -of_objects "itx/buf1/iBUF[$i].i_tri_buf_p/tri_buf"]
 }}
 set_dont_touch [get_nets "itx/buf1/BTN"]
 set_dont_touch [get_nets "itx/buf1/BTP"]
@@ -284,6 +286,10 @@ set_dont_touch [get_cells "itx/buf1/i_term_p"]
 # a combinational loop error
 set_false_path -through [get_pins -of_objects "itx/buf1/i_term_n"]
 set_false_path -through [get_pins -of_objects "itx/buf1/i_term_p"]
+
+# Make sure the transmitter is not retimed.  This may already be in
+# the main DC step, but it's not clear that it's being applied.
+set_dont_retime [get_cells itx]
 
 ######
 # MDLL
@@ -383,14 +389,14 @@ set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_0/din[0]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_0/din[1]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_0/din[2]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_0/din[3]}}]
-set_max_transition {0.00625*time_scale} [get_pin {{itx/qr_mux_4t1_0/data}}]
+set_max_transition {0.008*time_scale} [get_pin {{itx/qr_mux_4t1_0/data}}]
 
 # Mux -
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_1/din[0]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_1/din[1]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_1/din[2]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_1/din[3]}}]
-set_max_transition {0.00625*time_scale} [get_pin {{itx/qr_mux_4t1_1/data}}]
+set_max_transition {0.008*time_scale} [get_pin {{itx/qr_mux_4t1_1/data}}]
 
 echo [all_clocks]
 '''
--- a/designs/dragonphy_top/constraints/gen_constraints.py
+++ b/designs/dragonphy_top/constraints/gen_constraints.py
@@ -242,15 +242,19 @@ for {{set i 0}} {{$i < 4}} {{incr i}} {{
 # Input divider
 set_dont_touch [get_cells itx/indiv]
 
+# Internal nets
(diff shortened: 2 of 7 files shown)


designs/dragonphy_top/constraints/configure.yml
```diff
@@ -18,7 +18,7 @@ parameters:
 
   # Period of the main clock in nanoseconds
   # (will be scaled by constr_time_scale)
-  constr_main_per: 0.7
+  constr_main_per: 0.9
```
sgherbst, line 21: request from @zamyers

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -213,8 +213,6 @@
     iacore/*pi_out_meas* \\
     iacore/*del_out_rep* \\
     iacore/*inbuf_out_meas* \\
-    iacore/*pfd_inp_meas* \\
-    iacore/*pfd_inn_meas* \\
```
sgherbst, line 217: This was generating warnings, because these pins were removed awhile ago by @sjkim85.  I updated the constraints to reflect this.

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -233,6 +231,66 @@
 }}]
 set_dont_touch_network $tdbg_clk_pins
 
+# TODO: do we need to set dont_touch through the hierarchy?
```
sgherbst, line 234: This is where I set ``dont_touch`` on nets and cells in the transmitter.  I experimented with applying ``dont_touch`` at various levels, to nets, pins, and cells.  It may be that there is a simpler way to apply the constraint; I haven't tried to trim down the code yet.

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -283,12 +341,13 @@
...
-# Set transition time for high-speed signals monitored from iacore and itx
-# transition time is 10% of a 4 GHz period.
+# Set transition time for high-speed signals monitored from
+# iacore and itx.  transition time is 10% of a 4 GHz period.
 
-foreach x $adbg_clk_pins {{
-    set_max_transition {0.025*time_scale} $x
-}}
+# TODO: figure out how to set a transition constraint on a black-box pin
+# foreach x $adbg_clk_pins {{
```
sgherbst, line 348: This constraint didn't work, so I removed it.  Looks like applying ``max_transition`` to a black-box pin is problematic.  We still need to figure out a way to apply these constraints, though.

designs/dragonphy_top/constraints/gen_constraints.py
```diff
@@ -327,14 +389,14 @@
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_0/din[1]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_0/din[2]}}]
 set_max_transition {0.025*time_scale} [get_pin {{itx/qr_mux_4t1_0/din[3]}}]
-set_max_transition {0.00625*time_scale} [get_pin {{itx/qr_mux_4t1_0/data}}]
+set_max_transition {0.008*time_scale} [get_pin {{itx/qr_mux_4t1_0/data}}]
```
sgherbst, line 392: request from @zamyers (same with the other change to 0.008)

designs/dragonphy_top/qtm/termination.qtm.tcl
```diff
@@ -12,15 +12,21 @@ set_qtm_technology -library $::env(qtm_tech_lib)
...
- Vcm \
+    VinP \
+    VinN \
 }
 create_qtm_port $input_list -type input
 set_qtm_port_load -type qtm_load -factor 2 $input_list
 
+# define output(s)
+set output_list { \
+    Vcm \
```
sgherbst, line 25: The synthesis tool didn't like having an input driving an output (refused to wire it up).  It also didn't like having two ``inouts`` connected to a single output, so I settled on making ``Vcm`` an output and connecting it to ``DOUTP``/``DOUTN``.  It could be that making ``DOUTP``/``DOUTN`` ``inouts`` would work too, but I haven't tried it yet.

vlog/chip_src/tx_16t1/hr_16t4_mux_top.sv
```diff
@@ -14,7 +14,7 @@ module hr_16t4_mux_top (  // The output data rate should be input clock frequenc
 
 genvar i;
 generate  // Instantiate 4 hr_4t1_mux_top to form 16:4 mux
-    for (i=1; i<5; i=i+1) begin
+    for (i=1; i<5; i=i+1) begin : iMUX
```
sgherbst, line 17: ``generate`` blocks in code that is intended for synthesis should have block names, so that it is clear what the hierarchical path to blocks instantiated within the ``generate`` statement will be.

vlog/chip_src/tx_16t1/hr_2t1_mux_top.sv
```diff
@@ -10,16 +10,15 @@ module hr_2t1_mux_top (
...
 wire D0L; // din[0] wire connection from DFF to D-Latch
 wire D1M; // din[1] wire connection from DFF to MUX
 wire L0M; // din[0] wire connection from D-Latch to MUX
 
 //Instantiate the DFF, latch and MUX
 ff_c dff_0 (.D(din[0]), .CP(clk_b), .Q(D0L)); // DFF on din[0] path
 ff_c dff_1 (.D(din[1]), .CP(clk_b), .Q(D1M)); // DFF on din[0] path
-dlatch_n latch_0 (clk_b, D0L, L0M); // D-Latch after din[0] 
-mux mux_0 (L0M, D1M, clk_b, dout);
+dlatch_n latch_0 (.clk(clk_b), .din(D0L), .dout(L0M)); // D-Latch after din[0]
```
sgherbst, line 20: These modules had pins connected by order, not by name.  This is risky, because making a mistake on the pin order would not necessarily cause an error.  Plus, if the pin order changes for either of these modules, it would be more work to update it here.

vlog/chip_src/tx_16t1/qr_4t1_mux_top_2DFF.sv
```diff
@@ -14,25 +14,21 @@ module qr_4t1_mux_top_2DFF (
 
 // Instantiate the data path for Q clk path, use the Q clock as the reference clock
 wire D0DQ;
-wire D1MQ;
```
sgherbst, line 17: Several wires in ``qr_4t1_mux_top_2DFF.sv`` were unused (left over from the previous design iteration), so I deleted them to avoid confusion.

vlog/chip_src/tx_16t1/output_buf_tx.sv
```diff
@@ -52,15 +52,15 @@ module output_buf_tx (
     /////////////////
     
     termination i_term_n (
-        .Vcm(BTN),
-        .VinP(DOUTN),
-        .VinN(DOUTN)
+        .VinP(BTN),
```
sgherbst, line 55: I flipped the connections to the termination resistors, as mentioned earlier, so as to avoid a combo loop error.

CansWang (approved):
LGTM

PR #153, opened 2020-11-09
Constraints update

sgherbst:
## Summary

This PR contains constraints updates requested by @sjkim85 and @zamyers.

## Details
1. Add clock signals for all four input buffers (one had already been added), the ``clk_0`` MDLL output, and the TX input buffer.  The main purpose of this is to allow max transition constraints to be placed on those signals.
2. Fixed disconnection of TX termination resistors with update to constraints and I/O direction for the resistor black box.
3. Updated QTM for ``analog_core`` with new timing info from @sjkim85
4. Added max transition constraints for analog core debug clocks that are routed to the multiplexing output buffer in the digital core (all are treated as 4 GHz clocks)


sjkim85 (approved):
LGTM

PR #155, opened 2020-11-12
Fix missing output driver information

sgherbst:
We discovered that output driver information was missing on most pins in ``*.lib`` files generated with QTM.  It turns out the reason is that QTM will not add driver information to outputs that have no timing arcs.  A quick fix is to add a dummy arc from one of the QTM pins declared **clock** (not **input**) to the output(s) of interest.  This PR does exactly that, for all black-box pins that have a **create_clock** statement applied to them in the top-level synthesis constraints.

My main concern is that these dummy arcs might cause problems in the timing analysis -- please check the synthesis & PnR results to make sure that is not the case.  In parallel, I can investigate whether there is a cleaner way to add this driver information.  However, it is not clear that there will be, so I suggest that we use this quick fix for now.


sjkim85 (approved):
LGTM

PR #156, opened 2020-11-12
Phys design v2

zamyers:
1. Many changes to the build flow - including many covered by previous PR here.
   a. Added postroute_hold node to fix hold time issues after postroute - this node hasn't run yet so updates may be coming to debug this
   b. Added custom-route node to inject unique routing constraints into route step

2. Few logical fixes
   a. Removed a dead interface
   b. Fixed QR MUX to have correct output pin name


designs/dragonphy_top/constraints/configure.yml
```diff
@@ -18,7 +18,7 @@ parameters:
 
   # Period of the main clock in nanoseconds
   # (will be scaled by constr_time_scale)
-  constr_main_per: 0.9
+  constr_main_per: 0.7
```
sgherbst, line 21: Just wanted to check this is intended, since we had previously discussed increasing 0.7 to 0.9

dragonphy/__init__.py
```diff
@@ -22,7 +22,7 @@
 from .cmdparser		 import CmdLineParser
 from .packager       import Packager
 from .config 		 import Configuration
-from .tester     	 import DragonTester
+#from .tester     	 import DragonTester
```
sgherbst, line 25: Hmm, why is this commented out?  I think this would break a number of simulation-based tests.

setup.py
```diff
@@ -18,13 +18,13 @@
...
     'svinst==0.1.5',
     # magma ecosystem dependencies
-    'fault==3.0.36',
-    'magma-lang==2.1.17',
-    'coreir==2.0.120',
-    'mantle==2.0.10',
-    'hwtypes==1.4.3',
-    'ast_tools==0.0.30',
-    'kratos==0.0.31.1',
+#    'fault==3.0.36',
```
sgherbst, line 21: I think these were probably commented out accidentally and should be uncommented.

sgherbst (commented):
Generally looks good, just a few minor comments.  I think the reason the tests are failing is because some Python code got commented out in ``setup.py`` and ``__init__.py``

setup.py
zamyers: They aren't commented out accidentally. The server doesn't support these libraries and this was the fast solution to start physical design. They were accidentally committed however. 

(Tester is commented out because there is no magma/fault available).

sgherbst: OK, thanks for putting them back.  When there is more time post-tapeout, it would be great to find out more about this issue.  I imagine that the **magma** developers would like to support our design servers.

sgherbst (approved):
Great, thank you for submitting this PR!  I'll go ahead and merge it since the tests are passing.  Please be sure to keep the ``master`` branch current with the state of the design being used for tapeout.

PR #157, opened 2020-11-14
Add TX multicycle constraints

sgherbst:
This PR aims to resolve a hold time issue arising in CTS in the transmitter.  It appears that the default timing analysis of data being transferred from the half-rate clock to full-rate clock was not what we wanted, and I have attempted to fix that by using ``set_multicycle_path`` (which is really just a way of moving around the edges used for setup and hold checks).  I applied similar constraints to data being transferred from the quarter-rate clock to the half-rate clock.

@zamyers, please hold off on approving this PR until you can verify that the issue is resolved.  If more iteration is needed to solve the problem, I'll push those changes to this branch.  Thanks!


sgherbst:
Covered in #159

PR #158, opened 2020-11-14
Transmitter hot fix

CansWang:
Inverter added before 2GHz to 1GHz divider's input to fix the sampling timing.


sgherbst:
Covered in #159

